PR #72, opened 2017-07-13
changing deployment script to compile js

artemgurzhii:


mtlynch:
I use `script/_serve_dev.sh` to run jekyll locally. Can we update that as well in this PR?

PR #74, opened 2017-07-13
Changing newsletter sign up to be a hovering form

### LINKED ISSUE #52: Change newsletter sign up to be a hovering form
ConvertKit offers an option to have the newsletter form as a "slide in" that appears when the reader has scrolled X% down the page. I tried this initially, but it clashed with the site's CSS and didn't actually appear.

I'd like to switch to the slide-in style so that the form is more visible. It should slide in when the reader is 70% down the page in an article. The behavior on the article index page should stay the same (stay static on the bottom of the page).

artemgurzhii:
Fixes #52

### DIFF
--- a/_includes/convert_kit_article_form.html
+++ b/_includes/convert_kit_article_form.html
@@ -1,0 +1,220 @@
+<script src="https://assets.convertkit.com/assets/CKJS4.js?v=21"></script>
+<div class="ck_form_container ck_inline" data-ck-version="6">
+  <div class="ck_form ck_minimal" style="width: 100%; max-width: 100%;">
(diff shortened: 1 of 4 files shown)


mtlynch:
This looks good, but two notes:

* I'd like the behavior on the index page (like the https://mtlynch.io/ root) to stay the same and have a static sign-up form at the bottom:

image

* I don't want to change the form's text. I want it to have exactly the same text that it currently does.

mtlynch:
I tested this and I'm seeing the slide-in form on the index page:

image

On article pages, it is showing two forms:

image

PR #77, opened 2017-07-14
Travis deploy task

artemgurzhii:
Fixing travis-ci build script.


.travis.yml
```diff
@@ -11,6 +11,8 @@ script:
   - ./_tests/build
 before_deploy:
   # Build production site.
+  - npm install
+  - npm run build:js
```
mtlynch: These should go in `before_install` so that it runs on all PRs, not just deploys. #72 caused a build break because we couldn't see the build until the deploy step that happens after merge, whereas if it's in `before_install` it would have broke on the PR itself.

PR #82, opened 2017-07-20
ading class for the figure alignment

### LINKED ISSUE #79: Support an image + caption that aligns left or right
I'd like to have a tag or style that lets me create an image and caption with text wrapping around it.

I can currently do an image with no caption and have it align left with text wrapping around it:

image
https://github.com/mtlynch/mtlynch.io/blob/master/_posts/2017-05-07-building-a-vm-homelab.md#motherboard

And I can use the `<figure>` tag to create an image with a caption, but this takes up the full width of the content space:

image
https://github.com/mtlynch/mtlynch.io/blob/master/_posts/2017-06-27-greenpithumb.md#powered-by-raspberry-pi

I want something that lets me do both: align to the left or right of the page, but also have a caption for the image.

artemgurzhii:
Fixes #79

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -19,6 +19,10 @@ $screen-sm: 768px;
   p, li {
     font-size: 0.9em;
   }
+
+  div {
+    margin-bottom: 10px;
+  }
 }
 
 // Anchor Permalinks Component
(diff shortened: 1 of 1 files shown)


assets/css/main.scss
```diff
@@ -85,9 +85,26 @@ h1, h2, h3, h4, h5, h6 {
...
 
-
 .page__content {
   h1, h2, h3, h4, h5, h6 {
     margin: 0.5em 0;
   }
 }
+
+.figure-right-aligned,
+.figure-left-aligned {
```
mtlynch: Can we make this `figure.align-right` and `figure.align-left`? Otherwise the HTML looks repetitive:

```
<figure class="figure-right-aligned">
...
```

mtlynch:
It looks good. One issue is the caption should only be as wide as the image width. A long caption shouldn't cause the image to occupy the entire horizontal space:

image

PR #98, opened 2017-07-28
Feature: Customized Include Within Post

### LINKED ISSUE #94: Include a file in a post and allow user to download the raw file
In the [Sia Synology post](https://mtlynch.io/sia-via-docker/), I want to get rid of the `{% gist %}` snippet and host the file within the repo. So what I'd like is to do something like:

````
```bash
wget {{ site.url }}/files/2017-05-30-sia-via-docker_files/Dockerfile
```

The `Dockerfile` contents are listed below:

{% include 2017-05-30-sia-via-docker_files/Dockerfile %}
````

The problem is that my reading of the [docs](https://jekyllrb.com/docs/includes/) says that `{% include %}` is only for files in the `_includes` directory, which seems like a bad place for an article-specific file. And `{% include_relative %}` only allows files within `_posts` which actually seems doable, but then I don't know how to let the user download the raw file because non-Markdown files in `_posts` don't seem to get copied to the `_site` folder on build. I don't want to keep two copies of the file. I tried symlinking as a last resort and Jekyll didn't seem to like that.

Is there a good way of doing what I'm trying to do here?

### PR DESCRIPTION
**Pull request by shredtechular**
This PR Fixes #94.  This solution to the problem creates a Jekyll collection called "downloads" where we can then use a dedicated include file (_includes/downloads.html) that can be used to display the file within a Post/Page/Layout, etc.  There are other ways that this could probably be done, but I choose this solution because it meets the below requirements specifically:

- [X]  A content writer can simply use the standard include syntax in their markdown where they would like the file to be "embedded" in their Post. For example:  `{% include downloads.html title="XXX" %}`
- [X] The collection makes it possible so the file is also generated as is so it can be downloaded as well as included.
- [X] This solution has a reusability factor so that this functionality can exist in Posts, Pages, Layouts, other includes, etc.

## Open Questions about this solution
- [ ] Are we okay with having an `_downloads` directory that stores all types of gists/files-that-can-be-downloaded/etc?  Or did you want the URL to be relative to the Post or subdirectory?  Probably would need to redo the solution if so and maybe use js or something to inject the embed then, so this seemed like the best method.
- [ ] I titled the collection and directory "downloads" obviously, but we can make this something more semantic that may mean something more to you?  Gists?  Files?  Downloads made the most sense to me, but we can obviously change that if you have a better name.
- [ ] I included a "download raw" link underneath the `pre` block.  Did you want something standard for downloading files?  Or did you just want to be able to create a download link within the content itself?  If the former, then we may want to fancy up the download link some?  If the latter, then I can remove the download link within the include file and just leave it up to the content writer to add a download link
- [ ] Is there any styling that you need for including a file like this within a Post?  It obviously doesn't look exactly like the current gist embed, do you need it to?  I simply used `<pre>` to encompass it for the interim, however, if you envision this to be used simply for Gists then do you want any additional styling, line numbers, etc?  Below is how it looks locally:
screen shot 2017-07-27 at 9 00 08 pm

## Other Items of importance
- When adding a new file to the `downloads` collection, we do need to add yml frontmatter to the top of the file with a `title` field.  For example in the DockerFile, you will notice I put:
```
---
title: DockerFile
---
```
- I removed the `Jekyll-gist` gem from the _config.yml file as after a site wide search it seemed like the gist was only being used on the [Sia Synology post](https://mtlynch.io/sia-via-docker/).


mtlynch:
This is really great! Is there a way we can fix the compatibility with the build check that's currently failing?

Would it be possible to prefix the filename to make it unique per post, like `sia-via-docker/Dockerfile`? That way, I can have multiple posts that use a file named `Dockerfile`.

Also, be sure to include `Fixes #94` somewhere in the PR description so that when we merge this, it auto-closes the associated bug.

mtlynch:
Oh and good questions:

>I titled the collection and directory "downloads" obviously, but we can make this something more semantic that may mean something more to you?

Can we do `files`?

> I included a "download raw" link underneath the pre block. Did you want something standard for downloading files?

Download raw is a nice touch. I like that exactly as is.

>Is there any styling that you need for including a file like this within a Post

I think this is a good place to start and we can look at fancier things in the future if we need it.

shredtechular:
Thanks for the feedback @mtlynch !  I'm checking the build now, not sure why its failing.  The tests pass locally for me, but for some reason failing on Travis.  Gonna setup my own Travis instance so I can debug further, I'm sure I'm doing something stupid there...just gotta find it.

I'll edit the description to include "Fixes #94" once I get the build passing.

I'll also change the "downloads" references to "files", no problem there.

I'll see if I can prefix the filename to make it unique per post, like `sia-via-docker/Dockerfile`.  Pretty sure that will be doable, so I'll see what I can do there.

shredtechular:
OK, I found the issue with the build now and have fixed that as well as the below in commit: https://github.com/mtlynch/mtlynch.io/pull/98/commits/712e8eda0182d03cf475848dff252109cbeb4c86

- [X] Changed the collection name and all references from "downloads" to "files".
- [X] Made a directory within `_files` titled `sia-via-docker` and placed the DockerFile inside there.  This will allow the prepending of a filename (as long as it matches the directory name) to be included for a file and thus having multiple `DockerFile`s per Post/Page.

Finally, updated the description to include "Fixes #94", I'll be sure to do that moving forward on all PRs.

mtlynch:
This is awesome! Great work.

PR #104, opened 2017-08-01
Feature: Support for YML & dynamic syntax highlighting for custom embed include

### LINKED ISSUE #100: Support .yml files in _includes/files.html
#98 added support for embedding downloadable file contents in posts. It seems that Jekyll treats `.yml` files as a special case and doesn't display them properly in the `files.html` include. Is there a way to embed yaml files?

This is visible in a draft post (376de5b0a96464d15f17a15d8e1b2d0856de16be). Markdown is:

```
# Installation

`docker-compose.yml`

{% include files.html title="docker-compose.yml" %}
```

but the page renders as:

image

### LINKED ISSUE #101: Add support for non-bash syntax highlighting on files.html includes
#98 hardcodes the language of the embedded text to `bash`. It would be good if the embedded text defaults to a raw triple backtick (no language specified), but allows the caller to specify an optional language, e.g.

```
{% include files.html title="foo.cpp" language="cpp" %}
```

shredtechular:
This PR fixes #100.  In addition, this PR fixes #101.  The solution for #100 HAD to include the solution for #101 and they are not mutually exclusive for this solution.

In order to solve the issue for supporting yml files to be embedded within a post AS WELL AS downloadable as a yml file by an end user, the following was done:

A new Jekyll collection called "ymls" whose files are located in `_ymls` was created.  This collection is ONLY for yml files to be included within posts.  It works the exact same way as the "files" collection created in #98 however, what makes it different is its handling of how it is stored and then generated as detailed below.

- the "files" collection should include any and all files that are included within posts, with the exception of yml files, due to the special significance yml has for Jekyll.  The files collection will generate the files as is minus the yml frontmatter that is at the top of the file.
- the "ymls" collection should contain ONLY yml files that can not be include in "files" due to the special significance of yml for Jekyll.  The ymls collection will generate the files with a ".yml" extension.  That is really the only difference, but it allows support for yml files to do what we need to solve #100.
- in order to include yml files in posts, it uses the same liquid include syntax as the files collection includes, however, it *must* include the `language="yml"` parameter in the include as the logic keys on that parameter and checks if the language is yml in order to properly include it from the correct collection.

### What I like about this solution:
- It's native to Jekyll functionality
- We don't have to complicate build/deploy scripts by changing extensions and editing already generated html risking a bug of some sort
- It follows a similar methodology already used in #98.

### What I don't like about this solution:
- It adds another directory `_ymls` to the root for the Jekyll collection
- It separates out common functionality/storage which can add to complexity down the road

### Other minor notes in this PR:
- I added a "download" attribute to the "download raw" link in order to force the browser to actually download the file when clicked.  For yml files when I clicked a link in some browsers it just opened the yml file within the browser.  Adding the attribute forced the file to download.  Hope that is desired?  If not, we can remove the download attribute and let the browser decide how to handle.
- The syntax highlighting language will now default to null or raw triple backtick per #101 , unless the language parameter is stated in the include statement.
- I also kept one file in the `_ymls/` directory so the commit would keep the directory included in the commit and you wouldn't have to add that directory when you do add in yml files.  When this PR is merged, the yml file will be built into the `_site` dir but since it is not being linked anywhere in this commit/branch yet, that should be okay.

#### Non-yml files (doesn't have a `.yml` file extension)
1. place the code example file into the `_files` directory
2. edit the code example and add yml frontmatter with a title key, for example: 
    ```
    ---
    title: "filename.extension"
    ---
   ```
3. use the below syntax to include within a post with the `title` param *required* and matching the code example file's yml key for `title` and the `language` param *optional* signifying syntax highlighting language to be used:
    `{% include files.html title="filename.extension" language="bash" %}`

#### yml files
1. place example yml file into the `_ymls` directory
2. edit the example yml file adding yml frontmatter with a title key, for example: 
    ```
    ---
    title: "filename.yml"
    ---
   ```
3. use the below syntax to include within a post with the `title` param *required* and matching the code example file's yml key for `title` and the `language` param *required* and must be set to "yml":
    `{% include files.html title="filename.yml" language="yml" %}`

### DIFF
--- a/_includes/files.html
+++ b/_includes/files.html
@@ -7,22 +7,44 @@
   displayed inside a pre block and include a download link underneath.
 
   @param {string} "title" - the yml meta field for title in the download file
+  @param {string} "language" - the coding language used for syntax highlighting
 
 -->
 {% endcomment %}
 
+{% comment %}
+<!-- determine correct collection based on language param -->
+{% endcomment %}
+{% if include.language == "yml" or include.language == "yaml" %}
+	{% assign files = site.ymls | where: 'title', include.title %}
+{% else %}
+	{% assign files = site.files | where: 'title', include.title %}
+{% endif %}
 
-{% assign files = site.files | where: 'title', include.title %}
+{% comment %}
+<!-- determine syntax highlighting language.  Defaults to null -->
+{% endcomment %}
+{% assign lang_highlight = "" %}
+{% if include.language %}
+	{% assign lang_highlight = include.language %}
+{% endif %}
+
 {% for file in files %}
 
-  {% assign file_path = file.relative_path  | remove: '_files/' %}
+  {% comment %}
+  <!-- determine correct collection based on language param -->
+  {% endcomment %}
+  {% assign file_path = file.relative_path  | remove: '_files/'  | remove: '_ymls/' %}
+  {% if include.language == "yml" or include.language == "yaml" %}
+    {% assign file_path = file_path | append: "." | append: include.language %}
+  {% endif %}
   {% assign link = site.baseurl | append: '/files/' | append: file_path %}
 
-```bash
+```{{ lang_highlight }}
 {{ file.content }}
 ```
 <div class="wrapper">
-  <a href="{{ link }}" class="align-right">download raw</a>
+  <a href="{{ link }}" download class="align-right">download raw</a>
 </div>
 
 {% endfor %}
--- a/README.md
+++ b/README.md
@@ -70,3 +70,27 @@ If the PR fixes an issue, include the text "Fixes #XX" in the PR description, wh
 ### Oxford comma
 
 * Always use an Oxford comma.
+
+## Including code samples in posts
+
+### For non-yml files
+1. Place the code sample file into the `_files` directory
+2. Edit the code sample file and add yml frontmatter with a title key, for example: 
+   ```
+   ---
+   title: "filename.extension"
+   ---
+   ```
+3. In the post, use the below syntax to include the code sample.  The `title` param is *required* and needs to match the title key that was inserted in step 2. The `language` param is *optional* signifying what syntax highlighting language should be used.
+   `{% include files.html title="filename.extension" language="bash" %}`
+
+### For yml files
+1. Place the yml file into the `_ymls` directory
+2. Edit the yml file and add yml frontmatter with a title key, for example: 
+    ```
+    ---
+    title: "filename.yml"
+    ---
+   ```
+3. In the post, use the below syntax to include the yml file.  The `title` param is *required* and needs to match the title key that was inserted in step 2. The `language` param is also *required* and must be set to "yml"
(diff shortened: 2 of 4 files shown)


_includes/files.html
```diff
@@ -7,22 +7,44 @@
...
+<!-- determine correct collection based on language param -->
+{% endcomment %}
+{% if include.language == "yml" or include.language == "yaml" %}
+	{% assign files = site.ymls | where: 'title', include.title %}
+{% else %}
+	{% assign files = site.files | where: 'title', include.title %}
+{% endif %}
+
+{% comment %}
+<!-- determine syntax highlighting language.  Defaults to bash -->
```
mtlynch: Doesn't it just default to no highlighting language? Blank triple backticks is my desired behavior and it looks like that's what the code is doing, so it's probably just a matter of updating the comment.

mtlynch:
Oh, I like this. I agree that adding the `_yml` directory is a bit untidy, but it's much cleaner than what I thought we'd need to do.

Can we revert 376de5b? Otherwise, once I merge into `master`, Travis will deploy to my prod site with the in-progress NextCloud post.

_includes/files.html
shredtechular: Nice catch, you are exactly correct, the code is doing the correct thing.  My first attempt I did default it to bash, thus the comment, but after re-reading the issue description before submitting the PR I noticed I hadn't done that correct, so fixed the code, but forgot to update the comment.  Comment is updated in the newest commit now! 9dfb3b7322ce160cd280217d3059a8fcea35e59e

shredtechular:
Absolutely, I have completely backed out the commit and files related to the `NextCloud` branch.  I deleted my feature branch completely on my local system and re-branched off of `master`.  Then added the 2 necessary files/changes in for this PR.  So, just the single commit now without the `NextCloud` stuff.

I updated the PR description based on this change and would point out that I did keep one example yml file in the `_ymls` dir.  If I don't have any files in that directory, then it would be apart of the PR and you would have to create that dir separately when you are ready to add in yml files.  As that file isn't currently linked anywhere, I didn't see the danger in keeping an example file in there?

- [ ] One more question -- I also added some brief instructions for adding includable yml files and non-yml files into the PR description (titled "Instructions for including code examples in posts").  Do we want to add anything into `README.md` for this to have it documented somewhere about how to add includes?  Or does that get into the details too much and not needed?

mtlynch:
> Do we want to add anything into README.md for this to have it documented somewhere about how to add includes? Or does that get into the details too much and not needed?

Yeah, that's a good idea. That would be great!

shredtechular:
OK, I added some instructions to the README in 563ad7c7e46480c516ea0a67fa1f0015e032de20.  I added these instructions to the bottom of the README under a new H2 header titled "Including code samples in posts".  It felt like it should be it's own section to me...  Hopefully the instructions are clear!

mtlynch:
This is great! Thanks.

PR #109, opened 2017-08-06
Feature: Adding header hover link capability

### LINKED ISSUE #92: Add hoverlinks for headers
Headers in article content should have hoverlinks that allow the user to link directly to that heading. Headings outside of article content should not have hoverlinks.

This work is partially done in previous PRs, some of which have been reverted:

* #67 
* #72
* #77 

Issues with previous PRs:

* Any JS build step should happen in the `before_install` section of `.travis.yml` so that JS build failures show up in non-deployment builds
* The JS changes break image overlay functionality in blog posts. Normally, when the user clicks an image that has a larger size available (e.g. lots of examples in the GreenPiThumb post), it pops up an overlay of the image. When we do the `build:js` step, it breaks this functionality and sends the user away from the post to a bare image URL.
* The JS changes cause the newsletter sign-up form to block the page numbers in the article index (#86 is supposed to fix this, but I haven't confirmed this works on mobile)

shredtechular:
Fixes #92 

This PR is the easiest solution for adding the header hover link capability into the site.  This solution called for customizing the base Jekyll theme and seems like how it was intended to work from previous pull requests.

### Solution description

Adding the header hover links requires js to enable the GFM links that appear to the left of headers.  Due to the need of additional js to accomplish the task, we need a way to get the custom js into main.min.js which the site calls, therefore the build task is modified.  However, because we need to rebuild the js, we need to store all of the js from the base Jekyll theme in this local repo.  This may not be ideal from an updating gem perspective, but it is the quickest solution without having the site all additional js files.  Other options to avoid doing that are listed in "Other possible solutions to consider" below.
- [X] Extended the js capability by creating a separate js file (`assets/js/_anchors.js`) to house the anchor link capability.  Previous attempts at doing this completely over-rode the theme's default `_main.js` file so that is the reason for other functionality breaking like the image popups and incorrect styling issues with the sign-up form, etc.  The `assets/js/_main.js` file is copied directly from the theme (as it was for version 4.4.2 -- it has changed in newer versions of the theme's gem) without any modifications.

- [X] Added a feature which I believe is desired when clicking a hover link actually changing the url hash so it can be copy/pasted/linked.  Because of the base theme's implementation of JQuery smoothScroll, I had to override that plugin with how in-page anchors treat the hash.

- [X] Updated the `build:js` job in `package.json` to re-compile `main.min.js`.  Also updated the version of package.json to match the theme's gem being used.

- [X] Modified the build step in the `before_install` section of `.travis.yml` to re-compile the `main.min.js` file.  Also modified the development build script to include re-compiling the js just like revised previous attempts did.

- [X] Tested in Mac Chrome, FF, & Safari and several browser widths and Chrome's device emulator.  Having a hard time testing on a true mobile device and windows since the base theme uses absolute urls everywhere.

- [X] Tweaked the styling a bit as I saw issues with how the link was showing for different header types (h1, h2, h3, etc.).  I personally felt this styling looked better, but certainly happy to revert to previous styling if desired.

### Considerations in this solution

- [X] Previous attempts for this solution included es6 syntax for the js.  Keeping it like this is fine, however, that is the only es6 being used in the site/theme, so including babel dependencies for a comparatively small piece of the entire puzzle may want to be evaluated.  If there are future plans to use es6 more extensively, then maybe it makes sense to leave.

### Other possible solutions to consider
As mentioned previously, I'm not a big fan of copying all the js from the base theme when thinking about upgrading the gem to newer versions.  Overrides should be kept to a minimum in my opinion and limited to specially designated files.  I'm also not sure how often we need to upgrade the theme's gem in this site, but that's the whole point of using a gem.  Unfortunately, I couldn't come up with a solution where the site still had a single js file to load for this to work.
1. Another option, could be to build any custom js that we do on our end into its own js file and have the site load 2 js files?  I know best practice says to avoid that if at all possible for performance reasons, however, maybe this might be an exception to explore.
2. Along the same lines, I suppose we could inject a separate compiled js into an include/layout so the js is inlined, however, I feel the most efficient way of doing that is again overriding the theme.  So, I'd rather go with a separate js file, but mentioning it as a possibility.
3. I could also explore (not 100% positive this will work...) re-compiling the js post-build, but my concern is we wouldn't be able to have the js build in the `before_install` section in travis as it stands today. But would you be amenable to moving the Jekyll build to the `before_install`?  Or moving both commands to a `script` section?  
3. I also noticed in the theme's repo a request for adding in anchorjs and code was even submitted showing how.  Maybe in a future version of this theme, this functionality might exist in the base theme and we can remove/alter this solution if/when that is implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "minimal-mistakes",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "description": "Minimal Mistakes Jekyll theme npm build scripts",
   "repository": {
     "type": "git",
@@ -21,16 +21,13 @@
     "node": ">= 0.10.0"
   },
   "devDependencies": {
-    "babel-cli": "^6.24.1",
-    "babel-preset-es2015": "^6.24.1",
     "npm-run-all": "^1.7.0",
     "onchange": "^2.2.0",
     "uglify-js": "^2.6.1"
   },
   "scripts": {
-    "babel": "babel assets/js/_anchors.js --out-file assets/js/script-compiled.js",
-    "uglify": "uglifyjs assets/js/vendor/jquery/jquery-1.12.4.min.js assets/js/plugins/jquery.fitvids.js assets/js/plugins/jquery.greedy-navigation.js assets/js/plugins/jquery.magnific-popup.js assets/js/plugins/jquery.smooth-scroll.min.js assets/js/_main.js assets/js/script-compiled.js -c -m -o assets/js/main.min.js",
+    "uglify": "uglifyjs assets/js/vendor/jquery/jquery-1.12.4.min.js assets/js/plugins/jquery.fitvids.js assets/js/plugins/jquery.greedy-navigation.js assets/js/plugins/jquery.magnific-popup.js assets/js/plugins/jquery.smooth-scroll.min.js assets/js/_main.js assets/js/_anchors.js -c -m -o assets/js/main.min.js",
     "watch:js": "onchange \"assets/js/**/*.js\" -e \"assets/js/main.min.js\" -- npm run build:js",
-    "build:js": "npm run babel && npm run uglify"
+    "build:js": "npm run uglify"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "minimal-mistakes",
-  "version": "3.4.4",
+  "version": "4.4.2",
   "description": "Minimal Mistakes Jekyll theme npm build scripts",
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "minimal-mistakes",
-  "version": "3.4.4",
+  "version": "4.4.2",
```
mtlynch: Can we change to 4.5.0? I upgraded the gem while this was in flight.

mtlynch:
>Tested in Mac Chrome, FF, & Safari and several browser widths and Chrome's device emulator. Having a hard time testing on a true mobile device and windows since the base theme uses absolute urls everywhere.

I pushed to http://preview.mtlynch.io/ if you want to test it, but I checked on mobile and it looked good to me. Going forward, I should set you up with a GCS bucket you can push to for previews because it's pretty easy and will probably come in handy.

>Previous attempts for this solution included es6 syntax for the js. Keeping it like this is fine, however, that is the only es6 being used in the site/theme, so including babel dependencies for a comparatively small piece of the entire puzzle may want to be evaluated. If there are future plans to use es6 more extensively, then maybe it makes sense to leave.

No future plans for es6. The fewer dependencies the better. Does it make sense to pull that work into this PR or is that something we can do independently?

>Other possible solutions to consider

I like option (1). Options (2) and (3) feel too much like fighting with Jekyll or the theme itself and if (4) goes through, we can just drop our custom fix and switch to that.

mtlynch:
Oh, also this looks great!

shredtechular:
> No future plans for es6. The fewer dependencies the better. Does it make sense to pull that work into this PR or is that something we can do independently?

I went ahead and added commit 95bffb2 to remove the es6/babel dependencies from the project. 
 I also updated the version in `package.json` to 4.5.0 in this commit.  It seems like it makes sense to combine that commit with this PR... if you disagree, it is easily removable and can be a separate PR.

> I like option (1). Options (2) and (3) feel too much like fighting with Jekyll or the theme itself and if (4) goes through, we can just drop our custom fix and switch to that.

I should've researched option (2) more before I threw that out as an option because I ran into issues trying to get that option to work consistently.  Our custom js depends on JQuery, because the theme uses JQuery smoothScroll.  And unfortunately, having 2 separate scripts can cause a race condition if JQuery is not loaded first, not to mention, we still have to modify default theme files to even have a chance at having the 2 scripts load properly.  So, it sort of defeats the purpose of this option if we are still having to modify default theme files.

Hope that all makes sense and again apologize for not thinking that option through better to begin with.  So, I think the original solution is probably our best bet, and reading the theme's docs, I think this is actually how they did envision it.  Again, it's just too bad we have to keep duplicates of those files separate from the gem.

mtlynch:
Great, this looks good. Thanks!

PR #117, opened 2017-08-12
Removing leading zero for the day part of article dates

### LINKED ISSUE #96: Remove leading zero in day part of article dates
The blog article dates show up as:

May **07**, 2017

It should appear as:

May **7**, 2017

(no leading zero)

shredtechular:
This PR fixes #96.

In order to remove leading zeros from the day part of article dates, we need to tweak the date formatting slightly from `%d` for dates to `%-d` to signify no leading zeros. 
 
- [X] I ended up doing a search and replace across the site to ensure the date formatting is changed consistently across any dates.
- [X] I verified that dates on the article index page, individual article pages, as well as the "related posts" section are all no longer showing leading zeros in dates.

**Please note** - The theme does have the `%d` date format inside some of its other layouts that are not currently being used in the site.  Therefore, if other layouts are utilized within the site, we may need to alter the date formats in those other layouts.  I would not recommend pulling those other layouts down into the site at this moment simply to change the date formats as I don't think there are any plans to utilize any of the other layouts.  If you feel differently, we can do that though.

### DIFF
--- a/_includes/archive-single.html
+++ b/_includes/archive-single.html
@@ -31,7 +31,7 @@
       {% endif %}
     </h2>
     {% if post.date %}
-      <p class="page__date">{{ post.date | date: "%B %d, %Y" }}</p>
+      <p class="page__date">{{ post.date | date: "%B %-d, %Y" }}</p>
     {% endif %}
     {% if post.read_time %}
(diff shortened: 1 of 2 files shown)


mtlynch:
Great! Appreciate the thoroughness.

PR #126, opened 2017-08-21
Feature: Addition of jekyll-responsive-image plugin

### LINKED ISSUE #116: Research: Jekyll image plugins
The minimal mistakes 4.5.0 theme changed the width of the content area a bit. It used to be 764px wide, so I made 764px width copies of my images and then linked them to the full size image. But 4.5.0 seems to have expanded the width to 984px, so now all of my custom sizing looks weird:

image

I've seen some discussion of plugins that handle some of this image work for you automatically. A few examples (there may be more):

* https://github.com/wildlyinaccurate/jekyll-responsive-image
* https://github.com/robwierzbowski/jekyll-picture-tag
* https://github.com/netlify/jekyll-srcset

I'm interested in a solution that (in priority order):

* Minimizes the effort of adding an image to a post
* Adapts to changes in the theme or device
* Prevents users from downloading excessive data (e.g. downloading a 3440x1440 px image to be displayed in a 200px space).

shredtechular:
Fixes #116.

## Solution
This update adds the jekyll-responsive-image plugin into the site which provides the `srset` image capability for responsive images to attempt to have the browser decide which image is best for a user's browser.

- [X] The plugin's configurations are stored in the `responsive_image` key in the `_config` file.
- [X] The plugin generates different image files based on the `sizes` key in the plugin's configurations.  I added sizes based on common break points: 320, 480, 768, 1024.  If there are specific image sizes desired, they can be set there.
- [X] I added a `media_renditions` key that can be used for the `sizes` attribute in `` tags.  I based these on the CSS for `<article>` tags as the images are all within that container.  We can certainly tweak that as necessary too.
- [X] The plugin generates the different image files in an `images/resized` directory and then copies to `_site` on site generation.  I added the `images/resized` directory to .gitignore as I don't think we want these under source control?  However, adding them could speed up/cache those for builds.  The more images and sizes there are the longer builds will take, so something we may want to keep an eye on as more images are added to the plugin.
- [X] Once an image is placed in a `responsive_image` tag or `responsive_image_block` then it will generate the different image sizes.
- [X] I added one image to the plugin for an example in the `_posts/2017-07-25-editor.md` post.  If you don't want this example, I can remove. 
- [X] I configured the responsive image plugin template in `_includes/responsive-image.html` file.  I attempted to make this template as flexible with as many options as possible given the site's images.  For example, you can have an image in a `<figure>` tag, with a popover link, without a popover link, plain image tag, custom classes, etc.  If you think any of these options are unnecessary, we can modify.
- [X] Included a "prequel" type of include that will allow you to simply write `path: foo.png` within a post.  However, instead of directly calling the responsive image tags within a post, you will have to call the "prequel" include.  See below or in the `_posts/2017-7-25-editor.md` for an example:
## Example of calling the "prequel" image template in post:
```
{% include image.html file="foo.png" alt="foo alt" img_link=true fig_caption="foo fig caption" media_rendition="article" %}
```

## Additional notes
- [ ] I tested the template with a couple images that have the whitespace issues and this solution fixes the whitespace issues along with loading different size images based on browser/device.  I started to test this with other images as well that weren't full width and having the whitespace issues, such as those images that are aligned-left or aligned-right and the solution doesn't work for those since srcset bumps the image to full width.  So, if the intent is to use this for those images too, then we need more work on the template/solution.
- [ ] The plugin really is just re-scaling the images from what it looks like, so for the mobile images, you will probably notice that it looks very similar to the original image, just quite smaller.  An example before/after screen cap is below for your information:

### Before plugin:
<kbd>
screen shot 2017-08-20 at 9 57 41 pm
</kbd>

### After plugin:
<kbd>
screen shot 2017-08-20 at 9 57 59 pm
</kbd>

### DIFF
--- a/_includes/image.html
+++ b/_includes/image.html
@@ -12,10 +12,10 @@
   @param {boolean} "img_link" - Optional, default is false. If img should have
     the image overlay functionality that the theme provides.
   @param {string} "fig_caption" - Optional. Caption to be used with figure tag.
-  @param {string} "media_rendition" - Optional.  The title for the array of 
-    media queries to include for the "sizes" attribute on image tag that is 
-    listed in the global _config file. Currently, the only possible value is: 
-    "article", however, other arrays of sizes could be added in the future.
+  @param {string} "media_rendition" - Optional.  The title for the array of
+    media queries to include for the "sizes" attribute on image tag that is
+    listed in the global _config file. If it's not included then defaults to
+    "article".
 
 -->
 {% endcomment %}
@@ -27,10 +27,18 @@
 {% assign img_basepath = site.responsive_image.base_path | append: "/" | append: post_slug %}
 {% assign img_path = img_basepath | append: "/" | append: "full-traffic-graph.png" %}
 
+{% comment %}
+<!-- default media_rendition to "article" if not included in parmaters -->
+{% endcomment %}
+{% assign media_rendition = "article" %}
+{% if include.media_rendition %}
+  {% assign media_rendition = include.media_rendition %}
+{% endif %}
+
 {% responsive_image_block %}
     path: {{ img_path }}
     alt: {{ include.alt }}
     img_link: {{ include.img_link }}
     fig_caption: {{ include.fig_caption }}
-    media_rendition: {{ include.media_rendition }}
+    media_rendition: {{ media_rendition }}
 {% endresponsive_image_block %}
--- a/_includes/image.html
+++ b/_includes/image.html
@@ -25,7 +25,7 @@
 {% endcomment %}
 {% assign post_slug =  page.path | replace:'.html','' | replace: '.md','' | replace: '_posts/', '' %}
 {% assign img_basepath = site.responsive_image.base_path | append: "/" | append: post_slug %}
-{% assign img_path = img_basepath | append: "/" | append: "full-traffic-graph.png" %}
+{% assign img_path = img_basepath | append: "/" | append: include.file %}
 
 {% comment %}
 <!-- default media_rendition to "article" if not included in parmaters -->
--- a/_includes/image.html
+++ b/_includes/image.html
@@ -1,0 +1,36 @@
+{% comment %}
+<!--
+  Image Include
+
+  This partial is used as a "prequel" include that is used with the responsive
+  image template.  This partial's main purpose is to dynamically create the
+  full path for the image to then pass that path as well as the other
+  parameters to the responsive image plugin and template.
+
+  @param {string} "file" - Required. filename of the image.
+  @param {string} "alt" - Required. Alt attribute for image tag.
+  @param {boolean} "img_link" - Optional, default is false. If img should have
+    the image overlay functionality that the theme provides.
+  @param {string} "fig_caption" - Optional. Caption to be used with figure tag.
+  @param {string} "media_rendition" - Optional.  The title for the array of 
+    media queries to include for the "sizes" attribute on image tag that is 
+    listed in the global _config file. Currently, the only possible value is: 
+    "article", however, other arrays of sizes could be added in the future.
+
+-->
+{% endcomment %}
+
+{% comment %}
+<!-- determine the full path of the image -->
+{% endcomment %}
+{% assign post_slug =  page.path | replace:'.html','' | replace: '.md','' | replace: '_posts/', '' %}
+{% assign img_basepath = site.responsive_image.base_path | append: "/" | append: post_slug %}
+{% assign img_path = img_basepath | append: "/" | append: "full-traffic-graph.png" %}
+
+{% responsive_image_block %}
+    path: {{ img_path }}
+    alt: {{ include.alt }}
+    img_link: {{ include.img_link }}
+    fig_caption: {{ include.fig_caption }}
+    media_rendition: {{ include.media_rendition }}
+{% endresponsive_image_block %}
--- a/_includes/responsive-image.html
+++ b/_includes/responsive-image.html
@@ -1,0 +1,58 @@
+{% comment %}
+<!--
+  Responsive Image Include
+
+  This partial is used with the responsive image plugin.  In the _config file,
+  this template is used as the default for responsive images to configure a
+  srcset value for images to ensure the best option for an image is used for a
+  specific user's browser.
+
+  @param {string} "path" - Required. Path to image file relative to the root
+  @param {string} "alt" - Required. Alt tag for image
+  @param {boolean} "img_link" - Optional, default is false. If img should have
+    popover link
+  @param {boolean} "fig" - Optional, default is false. If image should be
+    inside a figure tag
+  @param {string} "fig_caption" - Optional. Caption to be used with figure tag
+  @param {string} "rendition" - Optional.  If image should have sizes options.
+
+-->
+{% endcomment %}
+
+{% comment %}
+<!-- determine all image options for srcset -->
+{% endcomment %}
+{% capture srcset %}
+    {% for i in resized %}
+        {{ site.url }}/{{ i.path }} {{ i.width }}w,
+    {% endfor %}
+     {{ site.url }}/{{ original.path }} {{ original.width }}w
+{% endcapture %}
+
+{% comment %}
+<!-- determine sizes configurations for image -->
+{% endcomment %}
+{% assign rendition= site.media_renditions.[media_rendition] %}
+{% if rendition != nil %}
+  {% assign sizes = rendition.sizes | join: ', ' %}
+{%endif%}
+
+{% if fig %}
+<figure>
+{% endif %}
+
+{% if img_link %}
+  <a href="{{ site.url }}/{{ original.path }}">
+    <img src="/{{ path }}" alt="{{ alt }}" srcset="{{ srcset | strip_newlines }}" sizes="{{ sizes }}" class="{{ class }}">
+  </a>
+{% else %}
+  <img src="/{{ path }}" alt="{{ alt }}" srcset="{{ srcset | strip_newlines }}" sizes="{{ sizes }}" class="{{ class }}">
+{% endif %}
+
+{% if fig_caption %}
+  <figcaption>{{ fig_caption }}</figcaption>
+{% endif %}
+
+{% if fig %}
+</figure>
+{% endif %}
--- a/_includes/responsive-image.html
+++ b/_includes/responsive-image.html
@@ -7,14 +7,15 @@
   srcset value for images to ensure the best option for an image is used for a
   specific user's browser.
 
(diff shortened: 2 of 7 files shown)


_includes/responsive-image.html
```diff
@@ -0,0 +1,58 @@
...
+<!--
+  Responsive Image Include
+
+  This partial is used with the responsive image plugin.  In the _config file,
+  this template is used as the default for responsive images to configure a
+  srcset value for images to ensure the best option for an image is used for a
+  specific user's browser.
+
+  @param {string} "path" - Required. Path to image file relative to the root
+  @param {string} "alt" - Required. Alt tag for image
```
mtlynch: tag -> attribute

_includes/responsive-image.html
```diff
@@ -0,0 +1,58 @@
+{% comment %}
+<!--
+  Responsive Image Include
+
+  This partial is used with the responsive image plugin.  In the _config file,
+  this template is used as the default for responsive images to configure a
+  srcset value for images to ensure the best option for an image is used for a
+  specific user's browser.
+
+  @param {string} "path" - Required. Path to image file relative to the root
```
mtlynch: Can we add periods to the ends of all the param explanations?

_includes/responsive-image.html
```diff
@@ -0,0 +1,58 @@
...
+  @param {string} "alt" - Required. Alt tag for image
+  @param {boolean} "img_link" - Optional, default is false. If img should have
+    popover link
```
mtlynch: Can we use other terms to explain this? I don't recognize the term "popover link."

_includes/responsive-image.html
```diff
@@ -0,0 +1,58 @@
...
+  @param {boolean} "fig" - Optional, default is false. If image should be
+    inside a figure tag
+  @param {string} "fig_caption" - Optional. Caption to be used with figure tag
```
mtlynch: Can we eliminate `fig` and assume that if `fig_caption` is set, it's meant to be a `<figure>`?

_includes/responsive-image.html
```diff
@@ -0,0 +1,58 @@
...
+  @param {string} "rendition" - Optional.  If image should have sizes options.
```
mtlynch: It's a `string` type, but the description sounds like a bool. What are the possible values for this? Also, is it `rendition` or `media_rendition`? It looks like it's getting called as `media_rendition`.

_posts/2017-07-25-editor.md
```diff
@@ -19,18 +19,16 @@ I started this blog in May of last year. I don't mean to brag, but by last April
...
-{% endcapture %}
-
-<figure>
-  {{ fig_img | markdownify | remove: "<p>" | remove: "</p>" }}
-  <figcaption>{{ fig_caption | markdownify | remove: "<p>" | remove: "</p>" }}</figcaption>
-</figure>
+{% assign fig_caption = "Number of unique readers visiting [mtlynch.io](https://mtlynch.io) per week" | markdownify | remove: "<p>" | remove: "</p>" %}
+
+{% responsive_image_block %}
+    path: images/2017-07-25-editor/full-traffic-graph.png
```
mtlynch: I think it's a pretty safe bet that all images will be within `images/` so can we assume paths are relative to that directory? 

Actually, is it possible to infer automatically that they'll be in `images/{{ filename w/o extension }}/foo.png`? It would be great if I could just write `path: foo.png` and it would know to find it in `images/2017-07-25-editor/foo.png`.

_includes/responsive-image.html
shredtechular: Absolutely, completed in f75ceb6

_includes/responsive-image.html
shredtechular: In f75ceb6a10573c87d7ef1dcd24219f59ebfec333 change to:

> Alt attribute for image tag.

_includes/responsive-image.html
shredtechular: In f75ceb6a10573c87d7ef1dcd24219f59ebfec333 changed to:

> Optional, default is false. If img should have the image overlay functionality that the theme provides.

_includes/responsive-image.html
shredtechular: Great points!  It is `media_rendition` nice catch.  Also updated the description in f75ceb6a10573c87d7ef1dcd24219f59ebfec333 to:

> The title for the array of media queries to include for the "sizes" attribute on image tag that is listed in the global _config file. Currently, the only possible value is: "article", however, other arrays of sizes could be added in the future.

_posts/2017-07-25-editor.md
shredtechular: Unfortunately, the plugin needs the full path of the image as it uses this to build the resized images.  Therefore, dynamically setting this in the responsive-image plugin template is not possible.

However, what I can offer as a solution and what I checked in under f75ceb6 is to create a "prequel" type of include located at `_includes/image.html`.  Having a "prequel" type of include will allow you to just write `path: foo.png` within a post.  However, instead of directly calling the responsive image tags within a post, you will have to call the "prequel" include.  See below or in the `_posts/2017-7-25-editor.md` for an example:

## Example of calling the "prequel" image template in post:
{% include image.html file="foo.png" alt="foo alt" img_link=true fig_caption="foo fig caption" media_rendition="article" %}

Hope that makes sense and I'll update the PR description with this change as well.  If you don't like this as a solution I'll remove it, but not sure how else to accomplish the ability to infer automatically the image path.

mtlynch:
>However, what I can offer as a solution and what I checked in under f75ceb6 is to create a "prequel" type of include located at _includes/image.html. Having a "prequel" type of include will allow you to simply write path: foo.png within a post. However, instead of directly calling the responsive image tags within a post, you will have to call the "prequel" include.

Oh, yeah, I like `_includes/image.html` except that it forces us to duplicate all of the documentation for our parameters. I imagine I'd just always use `_includes/image.html` instead of `_includes/responsive-image.html`. Would it be possible to merge the functionality into a single include called `_includes/image.html`?

Like we'd basically replace these lines:

https://github.com/shredtechular/mtlynch.io/blob/f75ceb6a10573c87d7ef1dcd24219f59ebfec333/_includes/image.html#L30-L36

with these lines:

https://github.com/shredtechular/mtlynch.io/blob/f75ceb6a10573c87d7ef1dcd24219f59ebfec333/_includes/responsive-image.html#L23-L59

_includes/responsive-image.html
```diff
@@ -7,14 +7,15 @@
...
-  @param {boolean} "fig" - Optional, default is false. If image should be
-    inside a figure tag
-  @param {string} "fig_caption" - Optional. Caption to be used with figure tag
-  @param {string} "rendition" - Optional.  If image should have sizes options.
+    the image overlay functionality that the theme provides.
+  @param {string} "fig_caption" - Optional. Caption to be used with figure tag.
+  @param {string} "media_rendition" - Optional.  The title for the array of 
+    media queries to include for the "sizes" attribute on image tag that is 
+    listed in the global _config file. Currently, the only possible value is: 
+    "article", however, other arrays of sizes could be added in the future.
```
mtlynch: Can we make `article` the default so that we can simplify to:

>{% include image.html file="full-traffic-graph.png" alt="Blog traffic graphs" img_link=true fig_caption=fig_caption ~~media_rendition="article"~~ %}

shredtechular: Absolutely, that is complete in 9d3584a.

shredtechular:
> Would it be possible to merge the functionality into a single include called _includes/image.html?

Unfortunately, that is not possible.  The plugin uses the `responsive_image_block` or `responsive_image` tag in order to determine what images to resize and it needs the full path to the image in that tag.

I understand that the current setup doesn't exactly follow a DRY methodology, so another option would be to simplify the `_includes/image.html` to purely create the full path to the image and that's it.  Then the `responsive_image_block` could be called from the post as well.  I didn't opt for that solution due to the fact it doubles the work you would need to do to insert an image.  It would look something like this in the post:
```
{% include image.html file="foo.png" %}
{% responsive_image_block %}
    path: {{ img_path }}
    alt: {{ include.alt }}
    img_link: {{ include.img_link }}
    fig_caption: {{ include.fig_caption }}
{% endresponsive_image_block %}
{% responsive_image path={{ img_path }} alt="Foo's alt" img_link=true fig_caption=fig_caption %}
```
instead of the single include currently in the post:
```
{% include image.html file="foo.png" alt="Foo's alt" img_link=true fig_caption=fig_caption %}
```

mtlynch:
>Unfortunately, that is not possible. The plugin uses the responsive_image_block or responsive_image tag in order to determine what images to resize and it needs the full path to the image in that tag.

Ah, gotcha. I realized I misunderstood. I thought `images.html` was including `responsive-image.html`, but I realize now it's the `responsive_image_block`, which is different.

_includes/image.html
```diff
@@ -0,0 +1,44 @@
...
+
+-->
+{% endcomment %}
+
+{% comment %}
+<!-- determine the full path of the image -->
+{% endcomment %}
+{% assign post_slug =  page.path | replace:'.html','' | replace: '.md','' | replace: '_posts/', '' %}
+{% assign img_basepath = site.responsive_image.base_path | append: "/" | append: post_slug %}
+{% assign img_path = img_basepath | append: "/" | append: "full-traffic-graph.png" %}
```
mtlynch: The `full-traffic-graph.png` looks like a copy/paste error

shredtechular: You are absolutely correct, that should not be in there...hard-coding mistake developing the solution.

That is fixed in c1e9aee.

shredtechular:
FYI, I was seeing some HTMLProofer failures sporadically yesterday on `http://amzn.to/2ohoJ3O` in the greenpithumb post.  Today, it seems like that URL is 404ing, so I added another ignore to the test build file in c1e9aee for this so Travis wouldn't error out.

mtlynch:
>FYI, I was seeing some HTMLProofer failures sporadically yesterday on http://amzn.to/2ohoJ3O in the greenpithumb post. Today, it seems like that URL is 404ing, so I added another ignore to the test build file in c1e9aee for this so Travis wouldn't error out.

Fixed in #127. That was a real failure so we actually don't want to suppress errors like that. With vimeo, it was 403'ing occasionally from Travis, but the link was actually fine. http://amzn.to/2ohoJ3O was completely broken as far as I could tell, so I generated a new copy.

I'll merge this and remove the suppression on my end but just something to be aware of in the future. I'll add something to the README about this.

mtlynch:
>I'll merge this and remove the suppression on my end but just something to be aware of in the future. I'll add something to the README about this.

Did this in #128.

PR #137, opened 2017-09-01
Feature: Migrating responsive images in /greenpithumb post

shredtechular:
This PR migrates the images in the /greenpithumb post over to using the jekyll-responsive-plugin implemented in #126.

In order to complete the image migration for the /greenpithumb post, all images were migrated into the usual image responsive tags within the post.  There were a couple of additional scenarios not previously accounted for in how images are displayed with the responsive plugin, so a few other changes in other files were required.  Those are itemized below:

- [X] the current /greenpithumb post has a couple of external image references.  The responsive image plugin will not be able to resize images if they are not local to the site.  Therefore, we can either (1) keep those images as external and not use the responsive image plugin style for those, or (2) download the images and store them in the codebase to use with the plugin.  I opted for option 2, but if you prefer the external images to stay external, we can change to option 1.

- [X] this post also contains scenarios of multiple images within 1 figure tag.  This required some additional `media_renditions` in the `_config.yml`.

- [X] finally, this post also had one example where the link around the image was not to be for an "image overlay" but is actually a link to an external page.  This was also not previously accounted for, so I altered the responsive plugin template to allow for that capability as well.


_includes/image.html
```diff
@@ -9,8 +9,11 @@
 
   @param {string} "file" - Required. filename of the image.
   @param {string} "alt" - Required. Alt attribute for image tag.
-  @param {boolean} "img_link" - Optional, default is false. If img should have
-    the image overlay functionality that the theme provides.
+  @param {string} "img_link" - Optional. The link when an image is clicked.
+    Possible values for img_link are:
+    - default is false or null. Means no link for the image.
+    - true. Means image will have overlay functionality that theme provides.
+    - an actual url.  Provides a page link to go to when image is clicked.
```
mtlynch: Can we make a separate parameter for URLs, like `link_url`? It feels a bit untidy to have overload `img_link` to where it can be null, true, false, or a string.

So it would be like:

```html
{% include image.html file="arduino-starter-kit.jpg" alt="Arduino starter kit" max_width="300px" class="align-left" link_url="http://amzn.to/2rpRLCf" %}
```

_posts/2017-06-27-greenpithumb.md
```diff
@@ -86,18 +64,9 @@ The code for GreenPiThumb is open-source and features:
...
-{% capture fig_caption %}
-GreenPiThumb wiring diagram ([downloadable file](https://github.com/JeetShetty/GreenPiThumb/tree/master/doc))
-{% endcapture %}
+{% assign fig_caption = "GreenPiThumb wiring diagram ([downloadable file](https://github.com/JeetShetty/GreenPiThumb/tree/master/doc))" | markdownify | remove: "<p>" | remove: "</p>" %}
 
-<figure>
-  {{ fig_img | markdownify | remove: "<p>" | remove: "</p>" }}
-  <figcaption>{{ fig_caption | markdownify | remove: "<p>" | remove: "</p>" }}</figcaption>
-</figure>
+{% include image.html file="greenpithumb_wiring.png" alt="GreenPiThumb wiring diagram" img_link="true" fig_caption=fig_caption %}
```
mtlynch: Can we actually keep this as the [external URL](https://raw.githubusercontent.com/JeetShetty/GreenPiThumb/master/doc/greenpithumb_wiring.png)? I'd prefer to just keep one authoritative copy of the wiring diagram.

_posts/2017-06-27-greenpithumb.md
```diff
@@ -107,18 +76,7 @@ Lastly, we have a [12V water pump](http://amzn.to/2p90wk8), but the Raspberry Pi
...
-
-{% capture fig_caption %}
-GreenPiThumb software architecture
-{% endcapture %}
-
-<figure>
-  {{ fig_img | markdownify | remove: "<p>" | remove: "</p>" }}
-  <figcaption>{{ fig_caption | markdownify | remove: "<p>" | remove: "</p>" }}</figcaption>
-</figure>
+{% include image.html file="greenpithumb_diagram.png" alt="GreenPiThumb software architecture" fig_caption="GreenPiThumb software architecture" %}
```
mtlynch: Ditto on this one. The advantage of using the docs.google.com URL is that the image updates instantly if I modify the Google Drawing.

_includes/image.html
shredtechular: Absolutely, I agree with you, that's a good point.  I changed the logic to look first for the `link_url` param and then the `img_url` param.  So, the possibilities are:

| param(s) included | rendering result |
| ------------------ | ---------------- |
| `link_url` | image will render with the link url value as the href in the `<a>` tag around the image |
| `img_url` | image will render with the image overlay capability |
| both `img_url` and `link_url` | image will use the `link_url` and  render with the link url value as the href in the `<a>` tag around the image |
| neither `img_url` or `link_url` | no link for the image |

This update is completed in 4903898

_posts/2017-06-27-greenpithumb.md
shredtechular: Absolutely, this has image has been reverted to the original external URL in 4903898

_posts/2017-06-27-greenpithumb.md
shredtechular: Absolutely, this has image has been reverted to the original external URL in 4903898

_posts/2017-06-27-greenpithumb.md
```diff
@@ -211,7 +231,7 @@ It's a lesson I learned [long ago](https://www.joelonsoftware.com/2002/03/04/not
 
 ## Electronics: start with the basics
 
-{% include image.html file="arduino-starter-kit.jpg" alt="Arduino starter kit" max_width="300px" class="align-left" img_link="<http://amzn.to/2rpRLCf>" %}
+{% include image.html file="arduino-starter-kit.jpg" alt="Arduino starter kit" max_width="300px" class="align-left" link_url="<http://amzn.to/2rpRLCf>" %}
```
mtlynch: Do we need the angle brackets or is it possible to just write the links as normal strings, e.g. `link_url="http://amzn.to/2rpRLCf"`?

shredtechular: Ah, yeah, I forgot to call that issue out.  If you don't include the angle brackets, then the markdown linter will fail on [Rule 034: No bare URLs allowed](https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md).  So, I added the angle brackets to work around that issue and that's how you would do it within markdown as well...  I know that isn't ideal or intuitive though...

The only way around the rule that I can think of would be:
1. Add an exception for that linting rule, but I don't really like or recommend that option.
2. We could configure the `link_url` parameter to just accept the url without the protocol.  For example, `link_url="amzn.to/2rpRLCf"` and then build the link within the template adding the `https` protocol.  That would fool the linter into not flagging that as an issue.  The only issue would be if you ever specifically would need regular `http`.  For example, it looks like the above example uses `http` vs `https` although it looks like the result goes to `https` either way, so not sure it matters in this case.  But if you did ever link to a URL that needed regular `http`, we would have to add yet another parameter to signify protocol if that were the case, which then starts to get a little clumsy in my opinion, but we can do something like that if you prefer instead of the angle brackets?

mtlynch: Oh, gotcha. You can turn off the linter rule. I don't use a lot of bare URLs in text, and I feel like when I do, I'll notice if Jekyll is not rendering it as a link.

shredtechular: OK, that makes sense and that'll work!  The linter rule for bare urls is now turned off in f4675fd so you no longer need to put angle brackets in with the `link_url` param either.

mtlynch:
Looks great, thanks!

PR #141, opened 2017-09-06
Feature: Migrating responsive images in /building-a-vm-homelab post

shredtechular:
This PR migrates the images in the /building-a-vm-homelab post over to using the jekyll-responsive-plugin implemented in #126.

This post contained 11 images and 8 of those were migrated to using the plugin.  The other 3 images were all external images.  They weren't gdoc or github images as were the case in previous PRs that had external images, but decided to keep them as external images for the initial PR.  If the desire is to make them local to the codebase, that can certainly be updated.  

Otherwise, this is a pretty straight forward migration and no unaccounted scenarios in this post.


mtlynch:
Looks great!

PR #142, opened 2017-09-06
Feature: Migrating responsive images in /taskrabbit-cooking post

shredtechular:
This PR migrates the images in the /taskrabbit-cooking post over to using the jekyll-responsive-plugin implemented in #126.

This post contained 4 images and all of them were migrated over to the plugin.

The only image that was slightly different was the last image in the post which originally has inline styles that include a grey border.  Instead of keeping the inline styles, I added a re-usable `img-border` class to the scss file which can then be included in the `class` parameter of the image.  

Otherwise, this is a straight forward conversion to the responsive image plugin.

**NOTE**:  I didn't include in this PR, but we may want to add a task if one doesn't already exist about organizing the custom CSS in the `assets/css/main.scss`.  Currently, I think any custom CSS is just being deposited in that file, but I have some other thoughts on how I think we can extend the theme's SASS in a more organized fashion instead of dropping all of the custom css into main.scss.  

This is probably a lower priority as there isn't that much custom CSS in that file, but should we add an issue/task for this?  Or is that what is meant by #84 perhaps?

### DIFF
--- a/_posts/2017-01-11-taskrabbit-cooking.md
+++ b/_posts/2017-01-11-taskrabbit-cooking.md
@@ -30,7 +30,7 @@ I thought this would be a good opportunity to experiment with outsourcing &mdash
 
 [TaskRabbit](https://www.taskrabbit.com/r/aig0/try) is a service that allows you to hire people for small tasks. They specialize in handyman type tasks, like home repair or furniture assembly, but the range of possible tasks is pretty broad. You can hire someone to stand in line for you, run your errands, or [attend birthday parties on your behalf](http://abcnews.go.com/Lifestyle/woman-hires-impersonator-attend-friends-birthday-party/story?id=29277677).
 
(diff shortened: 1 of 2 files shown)


mtlynch:
Looks good!

>NOTE: I didn't include in this PR, but we may want to add a task if one doesn't already exist about organizing the custom CSS in the assets/css/main.scss. Currently, I think any custom CSS is just being deposited in that file, but I have some other thoughts on how I think we can extend the theme's SASS in a more organized fashion instead of dropping all of the custom css into main.scss.

>This is probably a lower priority as there isn't that much custom CSS in that file, but should we add an issue/task for this? Or is that what is meant by #84 perhaps?

Oh, cool. I love refactoring, so I'm definitely interested in your ideas on how we can manage the CSS better. Can you file an issue about this to add more details about what you'd like to change?

#84 is just about adding a formatter to the build, like maybe something along the lines of [cssbeautify](https://www.npmjs.com/package/cssbeautify), but I haven't really finished fleshing out the idea, so it's just a stub for now.

Issue #143, opened 2017-09-06
Re-organize the site's local CSS to follow a traditional SASS/SCSS model


shredtechular:
Currently, all of the custom CSS that needs to be developed and extended outside of the parent theme's CSS has been stored in the `/assets/css/main.scss` file.  There is nothing wrong with this and it will work, however, if customizations to css were to continue, then eventually that main.scss file could become a bit untidy and possibly more difficult to maintain.  I think we can add a little more organization to the custom css by breaking it out into individual .scss files that are then compiled, along with the parent theme's .scss files, into the final main.css file that the site serves on the live website.

I originally and mistakenly understood that when building the site, the parent theme's _sass folder would completely overwrite anything we added to a local _sass directory and didn't think following a traditional SASS structure would be possible.  However, after a small experiment, I think is possible and could be beneficial from a maintainability perspective.

I would propose breaking out the `/assets/css/main.scss` file into .scss files within a `/_sass/child-theme/` directory.  In other words in the `/assets/css/main.scss` file, instead of all the custom css inside of that file, we would have a simple SASS include like:
```
@import "child-theme/main";
```
And then we could breakout all of the existing custom css into a SASS directory structure of .scss files, something similar to:
```
_sass
  |-> child-theme
     |-> base
        |-> _base.scss
     |-> components
        |-> _tabs.scss
     |-> helpers
        |-> _variables.scss
        |-> _mixins.scss
     main.scss
```

I think the main benefit here is really just maintainability and organization of CSS, so this doesn't need to be a really high priority, but just something to think about if it is worth it to do.

mtlynch:
Oh, cool. I like this idea. I've added it to the project queue.

PR #146, opened 2017-09-10
Feature: Convert subscribe form to MailChimp

### LINKED ISSUE #138: Switch mailing list form to MailChimp
The site currently uses ConvertKit for mailing list sign-ups, but I'd like to switch to MailChimp. I like the way the current form looks, though, so I'd like to keep the style pretty similar to what I have now, but posting to the MailChimp subscription backend. The style doesn't have to match exactly, but I'd like to use the same text and for the look to be as similar as possible:

```
Be the first to know when I post cool stuff
Subscribe to get my latest articles by email.
```

The existing partial is in `_includes/convert_kit_form.html`. I'd like to rename it to just `_includes/mailing_list_form.html` so we don't have to change the filename if we change providers.

Here is the MailChimp signup form for mtlynch.io:

```html
<!-- Begin MailChimp Signup Form -->
<link href="//cdn-images.mailchimp.com/embedcode/horizontal-slim-10_7.css" rel="stylesheet" type="text/css">
<style type="text/css">
	#mc_embed_signup{background:#fff; clear:left; font:14px Helvetica,Arial,sans-serif; width:100%;}
	/* Add your own MailChimp form style overrides in your site stylesheet or in this style block.
	   We recommend moving this block and the preceding CSS link to the HEAD of your HTML file. */
</style>
<div id="mc_embed_signup">
<form action="//mtlynch.us16.list-manage.com/subscribe/post?u=827c54a25e4f8abbb83d878c4&amp;id=7941b1a771" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate>
    <div id="mc_embed_signup_scroll">
	<label for="mce-EMAIL">Be the first to know when I post cool stuff</label>
	<input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="email address" required>
    <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
    <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_827c54a25e4f8abbb83d878c4_7941b1a771" tabindex="-1" value=""></div>
    <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button"></div>
    </div>
</form>
</div>

<!--End mc_embed_signup-->
```

shredtechular:
This PR fixes #138.

The changes included in this PR includes the bare minimum for converting the website's mailing list subscribe functionality from using ConvertKit to MailChimp.  Per the issue, this PR accomplishes the following requirements:

- [X] - It does keep the same design look and feel of the subscribe form (_other than a minor responsive change noted below..._)
- [X] - It does keep the same text/content for the form
- [X] - It does use MailChimp instead of ConvertKit

However, there were 2 things with the ConvertKit solution that weren't included with the MailChimp solution out of the box.  I wanted to get an initial PR together to review and to help illustrate the differences to begin with.  We can certainly optimize the issues as discussed below based on your feedback:

### MailChimp confirmation screen

The first difference that I see with the MailChimp embed form, is when a user submits a valid email address and clicks subscribe, a new tab is opened taking the user to the below confirmation screen:

screen shot 2017-09-10 at 1 13 01 pm

Is this expected or were we expecting similar functionality to ConvertKit for form submission?  The MailChimp confirmation page can be customized within the MailChimp settings for your account, like the content on the confirmation page and/or taking the user to a specific confirmation page within the local website.  

If this is not expected behavior, would you prefer that the confirmation/success message is all localized to the site?  In other words, don't take the user to the MailChimp default confirmation screen?  If so, I have seen a couple of options that we may be able to do, but wanted to confirm what you would like the most.  If you do want something more like the ConvertKit form submission confirmation, a couple of options I can explore below are:
1) We could possibly use their API to submit an AJAX call, however, their documentation makes it sound like this is not possible to do from the client and would need a server-side call.  I would recommend this option as I think their could be future benefits to utilize other endpoints of their REST API, however, I have a feeling this isn't going to work based on their current documentation. 

2) We could also cancel the supplied Mailchimp form action with js and mimic a form submission with AJAX.  The difference being it isn't using their API endpoint, but using the form action url to submit the information.  However, instead of using the default `form action` attribute, we use js to do this in the background.  This option should work and probably is the one we would have to go with.

### Form Validation

The second difference that I see is form validation out of the box.  The ConvertKit solution looks to have built in HTML5 browser validation (and maybe falls back to js as well).  The MailChimp embed form does not include any validation.  In fact, submitting a blank or invalid email results in the following MailChimp screen on form submission:

screen shot 2017-09-10 at 2 35 18 pm

Would you prefer any custom inline validation before form submission?  If so, that can be added, but wanted to check before spending the time.

## Other miscellaneous notes for current solution:
- [X] You may notice that I did slightly alter the default MailChimp structure/classes slightly to fit with current look and feel and took out inline styles, miscellaneous junk MailChimp puts in their embed.
- [X] I moved inline styles to the _main.scss file.  Again, more cleanup in this area will be done in #143 .
- [X] I also moved some of the actual text/content into the `_data/ui-text.yml` which includes other content across the site.  This way, if content tweaks need to be made in the future, no need to mess with the html template.  **NOTE** - I didn't add any other locale languages like Spanish, French, etc.  I don't think those are necessary/being used?
- [X] I did alter one (minor, I think?) thing with respect to how the form breaks down responsively.  With ConvertKit, the email field and subscribe button stack at < 435px browser widths, but I coded the MailChimp form to stack the email field and subscribe button to stack at < 600px which follows the parent theme's default breakpoints.

## Other Enhancement?
One other thing I was thinking about adding would be a "development" version of the form so local development/testing wouldn't add junk emails to your production subscription "list" within MailChimp.  In other words, we could add a configuration in `_config-dev.yml` to submit the form too when doing local development/testing.  This would make it so that local development would submit the form to a "development" list in MailChimp as opposed to your "production" list.  Not sure if you think this would be beneficial or not?  And does that make sense?  If you would like something like that, we just would need the embed for the development list I think.  Additionally, if we utilize their API we would just need a production `list_id` and a development `list_id`.

## Testing 
Tested this solution in:
- [X] Mac Chrome, FF, Safari
- [X] Iphone 5 Chrome, Safari
- [X] General responsiveness


mtlynch:
This looks good as-is. Thanks!

ConvertKit's experience is nicer in that you never have to leave the page, but MailChimp's experience is good enough for my needs. REST calls on the server side would be tough because there's no app server; the blog is all just static files running on Firebase. Maybe there's something possible with Firebase functions, but I'm not too worried about the new tab. Only like 0.01% of users sign up for the mailing list, so it's not a flow we have to optimize too heavily yet.

mtlynch:
Oh, forgot to respond to this note:

>One other thing I was thinking about adding would be a "development" version of the form so local development/testing wouldn't add junk emails to your production subscription "list" within MailChimp. In other words, we could add a configuration in _config-dev.yml to submit the form too when doing local development/testing. This would make it so that local development would submit the form to a "development" list in MailChimp as opposed to your "production" list. Not sure if you think this would be beneficial or not? And does that make sense? If you would like something like that, we just would need the embed for the development list I think. Additionally, if we utilize their API we would just need a production list_id and a development list_id.

I think this is a good idea, but we modify the mailing list so infrequently that I don't think it's worth adding a dev version yet. As long as we use names like "whatever.test@whatever.com" it's pretty easy for me to just purge those addresses from my email list once we're done with a round of development work.

PR #149, opened 2017-09-13
Feature: CSS fix for long captions on left & right aligned figure tags

### LINKED ISSUE #85: Handle long text in left or right aligned figure tags
Adding a very long caption to a left or right aligned image in a `<figure>` tag currently expands all the whitespace. I need a way of constraining the width of the whole element to just the width of the image.

image

shredtechular:
This PR fixes #85 

This update includes a CSS change that will ensure that the figure captions do not stretch past the width of a left or right aligned issue.  This should do the trick and does not adversely affect captions on figures that are NOT left or right aligned.

## Long caption example of left aligned figure after the fix:
<kbd>
screen shot 2017-09-13 at 5 56 03 pm
</kbd>

## Long caption example of right aligned figure after the fix:
<kbd>
screen shot 2017-09-13 at 5 54 29 pm
</kbd>

## Rare use case issue
There is one possible scenario where I think this solution may produce an undesirable result, however, I don't think it is very likely to occur so didn't take it into account.   That scenario could come into play if the figure caption contains a really long word that is larger than the image.  In this case, the image will stretch large enough to a width to contain that really long word.  As you can see in the below screen capture the "aaaa..." word extends the entire figure and breaks the rest of the caption to that minimum width as well:
<kbd>
screen shot 2017-09-13 at 6 13 04 pm
</kbd>

## Testing 
Tested this solution in:
- [X] Mac Chrome, FF, Safari
- [X] Iphone 5 Chrome, Safari
- [X] General responsiveness

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -95,6 +95,7 @@ figure {
     &-left {
       flex-direction: column;
       margin: 0 1em 0 0;
+      display: table;
 
       a {
         z-index: 1;
@@ -111,6 +112,11 @@ figure {
       margin-right: 1em;
     }
   }
+}
+
+figcaption {
+  display: table-caption;
+  caption-side: bottom;
 }
 
 // More robust clearfix class than what is in theme


mtlynch:
Excellent! Appreciate the thorough testing, as always.

PR #151, opened 2017-09-15
Feature: Reorganizing custom CSS into SASS model for child theme

### LINKED ISSUE #143: Re-organize the site's local CSS to follow a traditional SASS/SCSS model

shredtechular:
This PR fixes #143 .

This pull request reorganizes the local site's custom CSS to follow a traditional SASS structure for a website.  The benefits are more organized and reusable CSS to allow for easier maintenance of those custom styles.  This should also be an easier structure for future upgrades of the parent theme that could involve changes to the `assets/css/main.scss` file.

## Steps taken in solution:
- [X] all of the custom CSS that previously existed in `assets/css/main.scss` was moved to `_sass/child-theme` and was replaced by a simple `@import "child-theme/main"` statement to import of all the child theme's styles
- [X] the following directory structure was followed inside the `child-theme` SASS directory:
```
_sass
  |-> child-theme
     |-> base
        |-> _base.scss
        |-> etc...
     |-> components
        |-> _tabs.scss
        |-> etc...
     |-> helpers
        |-> _mixins.scss
        |-> etc...
     |-> layouts
        |-> _article.scss
        |-> etc...
     main.scss
```
- [X] the original custom CSS was then broken out into the defined structure, moving CSS to it's applicable SASS file.
- [X] some refactoring of obvious code that could be utilized using mixins was also completed.  There could be some additional areas we may want to refactor with variables and mixins, but I think we most want to get the custom CSS into the correct structure.

## Testing
- [X] Completed regression testing across all applicable browser to ensure no adverse affects to components dependent on child theme styles.


mtlynch:
This is great! I like the improved organization and modularization.

PR #164, opened 2017-10-01
Feature: CSS update for mobile floating figures

### LINKED ISSUE #161: Right aligned image causing ugly breaks
I'm working on a new post draft in the `code-reviews` branch. Preview is 

http://preview.mtlynch.io/human-code-reviews/

I noticed that on mobile, the right-aligned image causes really weird line breaks.

image

This might be something I have to be aware of when picking image sizes, but I wanted to check if there was anything that can be done in 

The ideal behavior would be that when there's so little horizontal space, the image would stop allowing text to float next to it and just expand to fill the space. But I realize that's probably hard to pull off, but I wanted to see if there are options that might make this scenario go smoother on mobile.

shredtechular:
This PR fixes #161.

This pull request contains CSS changes that will stop the floating of figures on mobile breakpoints to prevent ugly line breaks from occurring.

The reason that the ugly line breaks are occurring is that certain figures that "aligned left" or "aligned right" aren't big enough to demand 100% of the width so they allow a minimal amount of text alongside the image.  In cases pointed out in #161 that minimal width is reduced even more with the implementation of an unordered list element.

## Solution
I think the best solution here is to remove the "float" from figures on mobile breakpoints.  This is already done on plain images, but we had not implemented it yet for figures since we had not run into the ugly line break case to date for figures.

### Before the fix (new post)
<kbd>
screen shot 2017-10-01 at 11 31 55 am
</kbd>

### After the fix (new post)
<kbd>
screen shot 2017-10-01 at 11 16 04 am
</kbd>

This fix does affect other figure tags that are left or right aligned however.  Some of those figures didn't have an issue with line breaks on mobile previously, but this solution will adjust the figures for all situations.  If this is NOT desired, then I think we would need to apply a special class for the scenario.

### Example of another floated mobile image before fix
<kbd>
screen shot 2017-10-01 at 11 17 01 am
</kbd>

### Example of another floated mobile image after fix
<kbd>
screen shot 2017-10-01 at 11 17 10 am
</kbd>

### DIFF
--- a/_sass/child-theme/components/_figures.scss
+++ b/_sass/child-theme/components/_figures.scss
@@ -1,18 +1,24 @@
 // Figure tag classes
 
-figure {
-  &.align {
-    &-right,
-    &-left {
-      flex-direction: column;
-      margin: 0 1em 0 0;
-      display: table;
+figure.align {
+  &-right,
+  &-left {
 
-      a {
-        z-index: 1;
-      }
+    flex-direction: column;
+    margin: 0 auto;
+    display: table;
+    float: none;
+
+    @include breakpoint($small) {
+      margin: 0 1em 0 0;
     }
 
+    a {
+      z-index: 1;
+    }
+  }
+
+  @include breakpoint($small) {
     &-right {
       float: right;
       margin-left: 1em;


mtlynch:
Excellent, I like this fix!

PR #168, opened 2017-10-08
Feature: Centering images by default

### LINKED ISSUE #160: Permalinks not working on Chrome

### LINKED ISSUE #167: Bad caption alignment on center-aligned figure images
I noticed that when I use the following image include (in branch `cr1`):

```yaml
{% include image.html file="flowchart.jpg" alt="Code review flow" fig_caption="Placeholder chart, replace with illustration" max_width="600px" img_link=true %}
```

The caption appears to the right of the image, instead of below it:

image

I'd like for the include to generate an image that's limited to 600px, centered, and has the figure caption directly below the image.

shredtechular:
This issue fixes #167 

This update changes the default alignment for images and figures to centered.  Currently, the default is to left-align images, but this PR will change that to centering images and figures by default.

## Solution
In order to accomplish this, I added an override class in the child theme to center images and figures in the `base.scss` file.  In doing so, one feature that did break were images that were supposed to display side-by-side.  In order to fix that display issue, I did need to set the `display: flex` back for images that are supposed to align side-by-side with the `half` or `third` classes.  An example of this display is on the https://mtlynch.io/greenpithumb/ page.

Additionally, in order to keep things clean, I also removed all `align-center` classes within 3 posts.  `align-center` won't harm anything, but it also won't do anything since that is the default now.

## New HTMLProofer Ignore
Finally, I was receiving 403 errors on the upwork.com links with HTMLProofer when running the tests locally and in Travis.  I'm not sure if something has changed in their config or cert, but also added a new ignore for these URLs in the build script as well.

## Testing 
Tested this solution in:
- [X] Mac Chrome, FF, Safari
- [X] Iphone 6 Chrome, Safari

### DIFF
--- a/_posts/2016-11-26-prosperbot.md
+++ b/_posts/2016-11-26-prosperbot.md
@@ -36,7 +36,7 @@ To invest in peer to peer loans with Prosper, investors can use Prosper's web si
 
 I chose the API route and developed ProsperBot, a lending bot that continuously searches for new loans on Prosper and invests in them when they meet certain criteria. It also includes a web dashboard, which shows ProsperBot's current status and my Prosper account activity over time:
 
-{% include image.html file="prosperbot-frontend.png" alt="ProsperBot screenshot" img_link="true" max_width="700px" class="align-center img-border" %}
+{% include image.html file="prosperbot-frontend.png" alt="ProsperBot screenshot" img_link="true" max_width="700px" class="img-border" %}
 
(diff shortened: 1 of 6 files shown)


mtlynch:
Awesome! Sorry about the Upwork 403's. thanks for suppressing that.

PR #184, opened 2017-11-04
Bug: Removing link underlines in content

### LINKED ISSUE #182: Remove underlines from links
It looks like the latest theme upgrade added underlines to links

image

I like the no-underline look we had previously. Can we override the CSS for links so that we get rid of the underline?

shredtechular:
This PR fixes #182.

This update adds a css rule to the child theme SASS files for all `<a>` elements to remove the underline text decoration when the link is not being hovered.  I assume the desire would be to keep the underline in the hover state however?

## Testing 
Tested this solution in:
- [X] Mac Chrome, FF, Safari
- [X] Iphone 6 Chrome, Safari

Also tested to make sure that links in the header/footer/nav areas also remained unchanged as it seems as though those links kept the no-underline style.  The bug only seemed to affect links contained in content sections.  

**Please NOTE** this fix also removes the underlines on non "posts" pages as well, for example the projects page links no longer have underlines when in the non-hover state:

### Before
<kbd>
screen shot 2017-11-03 at 9 16 23 pm

</kbd>

### After
<kbd>
screen shot 2017-11-03 at 9 14 14 pm
</kbd>

### DIFF
--- a/_sass/child-theme/base/_base.scss
+++ b/_sass/child-theme/base/_base.scss
@@ -13,3 +13,7 @@ img {
   display: block;
   margin: 0 auto;
 }
+
+a {
+  text-decoration: none;
+}


mtlynch:
Excellent, thanks!

PR #193, opened 2017-11-11
Bug: Ipad display for blog posts fix

### LINKED ISSUE #192: Bug: Ipad display issue for blog posts
During some recent testing, I noticed a display issue on landscape views for tablets.  I know this affects ipads and probably other devices with similar browser widths.  I will submit a fix for this, but also wanted to document quickly as an issue for reference/tracking.

Below is a screen-cap of what I see on ipad (landscape), as you can see the actual article content is pushed down as there is a space overlapping issue between the sidebar and article itself.

<kbd>
screen shot 2017-11-11 at 4 43 16 pm
</kbd>

shredtechular:
This PR fixes #192 .

This code change contains a simple CSS change that will slightly modify the width of an article so that blog posts on landscape views of tablets (specifically ipads) will not drop down below the left sidebar on initial loading of the article at the top of the page.

Probably the best way to explain this is to show a before the fix (current production) and after the fix screenshot:

## Before the fix contained in this PR:
<kbd>

</kbd> 

## With the fix contained in this PR:
<kbd>

</kbd>

### DIFF
--- a/_sass/child-theme/layouts/_page.scss
+++ b/_sass/child-theme/layouts/_page.scss
@@ -2,7 +2,7 @@
 
 .page {
   @include breakpoint($large) {
-  width: span(10 of 12);
+  width: span(9.5 of 12);
   padding-left: gutter(0.5 of 12);
   padding-right: gutter(2 of 12);
 }


mtlynch:
Good catch! Thanks for fixing this.

PR #200, opened 2017-12-05
Bug: Fixing figure and image tag spacing to be equivalent

### LINKED ISSUE #199: Reduce spacing on <figure> posts
I noticed that `<figure>` tags create a lot of vertical whitespace as opposed to normal image includes.

## Normal image

From ["How to Do Code Reviews Like a Human (Part Two)"](https://mtlynch.io/human-code-reviews-2/).

image-whitespace

## Figure tag

From ["Hiring an Editor Grew My Blog Audience 450x"](https://mtlynch.io/editor/)

figure-whitespace

Would it be possible to adjust the spacing so that both types of images have roughly the same amount of vertical spacing?

shredtechular:
This PR fixes #199.

As pointed out in #199 there was more top/bottom margin spacing on `<figure>` tags as opposed to `<image>` tags.  In order to make the spacing (margins) equivalent, this required a few minor CSS rules to be applied to `<figure>` and `<figcaption>` tags.  The rules in this PR make the spacing for the tags exactly the same now.

As you can see in the screen caps below, the margins are now the same for above/below `<image>` tags as well as `<figure>` tags:

### Before bugfix
As you can see the current site has 44px computed margins above and below `figure` tags.
<kbd>
screen shot 2017-12-04 at 9 29 24 pm
</kbd>

And on `<image>` tags, currently the margins are based on the parent `<p>` tag and has a computed margin of 25.74px above and below.
<kbd>
screen shot 2017-12-04 at 9 42 54 pm
</kbd>

### After bugfix
With the code contained in this bugfix, you can see the computed margins for `<figure>` tags are now 0 on top and 25.74px on the bottom.  So, the bottom margins are now exactly the same.
<kbd>
screen shot 2017-12-04 at 9 28 36 pm
</kbd>

Finally, you can see that above the figures are going to be a `<p>` tag which has a computed bottom margin of 25.74px already, so now the top margins match exactly just like the current `<image>` tags.
<kbd>
screen shot 2017-12-04 at 9 28 41 pm
</kbd>

### DIFF
--- a/_sass/child-theme/base/_base.scss
+++ b/_sass/child-theme/base/_base.scss
@@ -6,7 +6,8 @@ h1, h2, h3, h4, h5, h6 {
 
 figure {
   display: table;
-  margin: 2em auto;
+  margin: 0em auto 1.3em;
+  font-size: 0.9em;
 }
 
 img {
--- a/_sass/child-theme/components/_figures.scss
+++ b/_sass/child-theme/components/_figures.scss
@@ -38,6 +38,7 @@ figure.align {
 figcaption {
   display: table-caption;
   caption-side: bottom;
+  margin-bottom: 0;
 }
 
 .img-border {


mtlynch:
Awesome, thanks!

PR #209, opened 2017-12-31
Bug: Firefox image max widths fix

### LINKED ISSUE #205: max_width should still apply when fig_caption is set
I notice that when I apply `fig_caption` to an image with `max_width`, the `max_width` parameter is ignored.

For example, in [this post](https://github.com/mtlynch/mtlynch.io/blob/bc74167488587d55d959c7f97a8821d903279def/_posts/2017-11-14-how-to-hire-a-cartoonist-for-your-blog.md), I have this snippet:

```
{% assign fig_caption = "Cartoon submission from Sofia" | markdownify | remove: "<p>" | remove: "</p>" %}

{% include image.html file="sample-sofia.jpg" max_width="500px" alt="Sample submission from Sofia" img_link=true fig_caption=fig_caption %}
```

Which results in this image:

image

But the width is 768px instead of 500px. Is it possible to make the image match the `max_width` parameter?

shredtechular:
This PR fixes #205.

Interestingly enough, it appears doing some testing on FF, Chrome, Safari, and IE Edge that this issue only affects Firefox.  It appears as though Firefox is the lone browser that is allowing CSS to override the `sizes` attribute of image tags.

So, my solution will force inline CSS to override the default CSS if a max-width is being set.  This appears to set the width properly in Firefox and doesn't impact other browser's rendering of the image either.  That's the best thing I can think to offer here since Firefox is rendering differently...

Here is the image displaying with the inline CSS fix on the `cartoonists` branch:

screen shot 2017-12-30 at 7 56 18 pm

### DIFF
--- a/_includes/responsive-image.html
+++ b/_includes/responsive-image.html
@@ -67,10 +67,10 @@
 
 {% if img_url %}
 <a href="{{ img_url }}">
-  <img src="/{{ path }}" alt="{{ alt }}" srcset="{{ srcset | strip_newlines }}" sizes="{{ sizes }}" class="{{ class }}">
+  <img src="/{{ path }}" alt="{{ alt }}" srcset="{{ srcset | strip_newlines }}" sizes="{{ sizes }}" class="{{ class }}"{% if max_width %} style="max-width:{{ sizes }}"{% endif %}>
 </a>
 {% else %}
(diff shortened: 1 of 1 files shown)


mtlynch:
Thanks!

Issue #176, opened 2017-10-17
Link author image and name to blog root


mtlynch:
I'd like the author name and name to appear the same as they currently appear, but to link to the site's root.

Can we see if we can apply this PR directly in the [minimal mistakes repo](https://github.com/mmistakes/minimal-mistakes)?

PR #213, opened 2018-01-06
Feature: Upgrading Minimal Mistakes theme to 4.9.0

### LINKED ISSUE #211: Upgrade to minimal mistakes 4.9.0
Looks like some good updates, including @shredtechular's fix to #176.

shredtechular:
This PR fixes #211 and #176.

The updates included in this pull request will bump the parent theme (Minimal Mistakes) from version 4.6.0 to 4.9.0 which as stated in #211 includes the fixes for #176.

In order to update the parent theme the following items needed to be done:
- [X] updating the gem version in the `Gemfile` and `package.json`
- [X] bumping the jekyll version to the latest of `3.7.0` (the parent them's minimum jekyll version is `3.6.0` but figure might as well bump to latest...)
- [X] setting the author `home: \` key in `_config.yml` which will link the author name & avatar to the root of the website, per #176.
- [X] finally, the theme also bumped to the next major version of FontAwesome, so per guidance from the theme followed [these directions](https://fontawesome.com/how-to-use/upgrading-from-4) which basically means we needed to adjust the font awesome icon classes in several of the layouts/includes too.
- [X] performed light regression testing in each of the major current browsers and devices.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
-gem 'jekyll', '~> 3.5.1'
-gem 'minimal-mistakes-jekyll', '4.6.0'
+gem 'jekyll', '~> 3.7.0'
+gem 'minimal-mistakes-jekyll', '4.9.0'
 gem 'jekyll-admin', group: :jekyll_plugins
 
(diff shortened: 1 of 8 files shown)


mtlynch:
Thanks!

PR #216, opened 2018-01-09
 Feature: Updating theme to 4.9.0 -- take 2

### LINKED ISSUE #211: Upgrade to minimal mistakes 4.9.0

shredtechular:
This PR fixes #211.

Unfortunately #213 was not a successful upgrade of the parent theme due to the google analytics snippet not being included in the building of the generated HTML files.

After some digging and troubleshooting, I discovered that the parent theme included updates to [disable analytics and comments in development environments](https://github.com/mmistakes/minimal-mistakes/releases/tag/4.8.0) in version 4.8.0.  In the current `_tests/build` there was no `JEKYLL_ENV` variable being set so the builds were defaulting to "development" by default.  This PR now includes an update to that file to ensure the production build is set with the `JEKYLL_ENV=production`.

Everything else in the previous PR #213 was accurate, we just needed the environment variable to be set for production.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
-gem 'jekyll', '~> 3.5.1'
-gem 'minimal-mistakes-jekyll', '4.6.0'
+gem 'jekyll', '~> 3.7.0'
+gem 'minimal-mistakes-jekyll', '4.9.0'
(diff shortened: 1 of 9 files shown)


mtlynch:
Ah, didn't realize we were building a dev version. Thanks!

PR #226, opened 2018-02-06
Feature: Migrating build process to docker container (Locale fixed)

### LINKED ISSUE #221: Run build steps in Docker container
I'd like to change the Travis config to use Docker to run the build tests and produce the production site files.

~~There is a similar example of this in one of my other projects (building a static site with Angular rather than Jekyll). The particulars are different, but we want to use the same general structure:~~

1. ~~The [Dockerfile](https://github.com/mtlynch/ketohub_web_app/blob/master/Dockerfile) copies the local source folder into the Docker image's `/app` directory.~~
1. ~~The [Dockerfile](https://github.com/mtlynch/ketohub_web_app/blob/master/Dockerfile) runs all the build tests.~~
1. ~~The [Dockerfile](https://github.com/mtlynch/ketohub_web_app/blob/master/Dockerfile) ultimately creates a production build by running the npm [build script](https://github.com/mtlynch/ketohub_web_app/blob/1eabed495eb67c240b6cb65e01d5a4cec92814d8/package.json#L7)~~
1. ~~The [.travis.yml](https://github.com/mtlynch/ketohub_web_app/blob/ab5af15920da64a512aead7fdcd414645bf566ec/.travis.yml#L12-L13) builds and runs the Docker image defined above~~
1. ~~The [.travis.yml](https://github.com/mtlynch/ketohub_web_app/blob/ab5af15920da64a512aead7fdcd414645bf566ec/.travis.yml#L14) copies the Docker container's `/app/dist` folder (which contains the production build files) to the host's `./dist` folder.~~

Instead of the [node:8-stretch](https://github.com/mtlynch/ketohub_web_app/blob/815fdba2c510ddcfc350bd4e580d058b6e1435f6/Dockerfile#L1) base image, we'll use the [jekyll/jekyll](https://github.com/envygeeks/jekyll-docker/blob/master/README.md) Docker image.

~~Then we would get rid of the existing [build file](https://github.com/mtlynch/mtlynch.io/blob/master/_tests/build) and move those steps to a Dockerfile, like in the ketohub_web_app example. We can get rid of the `--quick` flag because it's not worth trying to reimplement that in a Dockerfile.~~

**Update**: I think I have a better design. See below

shredtechular:
This PR builds off of #224 and fixes #221.

As described in #221 the goal of this feature is to move the current build and test process to a Docker container. @mtlynch recreated this process completely in #224 however the node:8-stretch doesn't set the locale in its image that I could find so I added locale settings in c09bbd4 to the `Dockerfile` and that resolves the Unicode errors experienced.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,5 @@
 sudo: false
-language: ruby
-rvm:
-  - 2.4.1
-env:
-  global:
-    - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
-before_install:
-  - nvm install 7.10.0
-  - npm install
-  - npm run build:js
-  # Remove Jekyll Admin from Travis builds.
-  - sed -i '/jekyll-admin/d' Gemfile
-script:
-  - ./_tests/build
+script: pushd _tests && ./docker_build && popd
 deploy:
   provider: firebase
   token:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 services:
   - docker
 dist: trusty
-script: pushd _tests && ./docker_build && popd
+script: _tests/docker_build
 deploy:
   provider: firebase
   token:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
-sudo: false
+sudo: required
+services:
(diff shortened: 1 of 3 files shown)


mtlynch:
Ah, awesome!

PR #246, opened 2018-03-27
Making Homelab remote images local

### LINKED ISSUE #245: Images too large in homelab post
At some point in theme adjustments, some of the images in the homelab post got screwed up:

image

The problem seems to be that for these images, we're using hotlinks instead of storing them locally and using the image.html include. 

```
[![Samsung 850 EVO](//ws-na.amazon-adsystem.com/widgets/q?_encoding=UTF8&ASIN=B00TGIVZTW&Format=_SL250_&ID=AsinImage&MarketPlace=US&ServiceVersion=20070822&WS=1&tag=mtlynch-20){: .align-left}](https://www.amazon.com/gp/product/B00TGIVZTW/ref=as_li_ss_il?ie=UTF8&psc=1&linkCode=li2&tag=mtlynch-20&linkId=5406b1b2957a1e42f41c407ba8fd6559)
```

We should just switch to the image include and set the max width to the image's actual width.

shredtechular:
This PR fixes #245.

I agree, I think the best path to fixing the remote image sizing in the homelab post is making the images local and using the responsive image plugin for sizing.  It adds a level of control to the site in how they display and are sized.  And they appear to be very specific images of products that aren't just going to change.  Not to mention, these images should all be optimized for responsiveness too.

So, I pulled those 3 images in the homelab post locally and put them into the image include tags.  It continue to uses the same alt tags, links, and class alignments.  The only change is where the image file reference is now basically.

I would point out that in searching for other remote images within the site's markdown files, I found one other image that is currently pointing to a remote image in the GreenPiThumb post here: [https://mtlynch.io/greenpithumb/#hardware-architecture](https://mtlynch.io/greenpithumb/#hardware-architecture).  I did NOT switch this image as I thought I remember the logic being then maintaining the image in 2 places, so I held off making this image local.  If that is desired, I can certainly do the same for that image.


mtlynch:
Excellent, thanks!

>I would point out that in searching for other remote images within the site's markdown files, I found one other image that is currently pointing to a remote image in the GreenPiThumb post here: https://mtlynch.io/greenpithumb/#hardware-architecture. I did NOT switch this image as I thought I remember the logic being then maintaining the image in 2 places, so I held off making this image local. If that is desired, I can certainly do the same for that image.

Yep, that's correct. Let's leave that one as is. I control the source for that one so I'm more comfortable hotlinking it.

PR #266, opened 2018-06-09
Feature: Adding post frontmatter build checks

### LINKED ISSUE #227: Add build checks for frontmatter fields in posts
I'd like to add a build check to make sure every post has the following frontmatter fields defined:

* `header.og_image`
* `excerpt`

`header.og_image` can be defined to `null`, but it should still be defined. Can we distinguish between undefined and `null`?

shredtechular:
This PR fixes #227 .

This update adds a Jekyll pre-render hook that will run every time a `jekyll serve` or `jekyll build` occurs on the site and check frontmatter values against a schema defined in `_linter/frontmatter-tests/_posts.yml`.  

We could simply add some additional grep tests into the `_tests/build` file that I think that may satisfy the requirement for #227.  However, I was thinking a better solution would be to create something a little more flexible against a schema that we could hopefully extend to other frontmatter fields in the future if need be as well as have these checks run anytime and in every environment the site is built.  I think some of the checks curently in `_tests/build` are best run only for production builds and would slow down local development and local editing.  So, that is why I was thinking a local Jekyll pre-render plugin check for the solution to this requirement.

As far as differentiating between `null` and undefined -- these checks will **FAIL** if:
- either `header.og-image` or `excerpt` is not in the frontmatter at all of each `_post/xxx.md` file
- either either of those fields are set to "undefined", for example: `excerpt: ` or `og-image: `
If the fields are set to `null`, the checks will **PASS** for example:
- `og-image: ""` 
- `excerpt: ""`
So, with empty double quotes the values pass as `null`.  If that isn't correct or desired, we can change that.

Finally, there were a number of Posts that didn't have both of these fields set, so in order to get the build to pass, I added `og-image: ""` to any post that didn't have `og-image` set and `excerpt: "whatever the current excerpt is on the site"` for posts that didn't have `excerpt` set.

Lastly, is a screenshot of how the checks fail and look before I fixed the Posts adding the fields:

screen shot 2018-06-09 at 4 40 28 pm

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/mtlynch.io/266)

### DIFF
--- a/_linter/frontmatter-tests/_posts.yml
+++ b/_linter/frontmatter-tests/_posts.yml
@@ -1,0 +1,15 @@
+---
+excerpt: "String"
+header: 
+  type: "Dictionary"
+  key: "og_image"
+  key-type: "String"
+config:
+  path: _posts
+  ignore:
+  - README.md
+  - ..
+  - .
+  - .DS_Store
+  optional: ['']
+---
--- a/_linter/frontmatter-tests/_posts.yml
+++ b/_linter/frontmatter-tests/_posts.yml
@@ -1,6 +1,6 @@
 ---
 excerpt: "String"
-header: 
+header:
   type: "Dictionary"
   key: "og_image"
   key-type: "String"
--- a/_plugins/jekyll_frontmatter_tests.rb
+++ b/_plugins/jekyll_frontmatter_tests.rb
@@ -1,0 +1,179 @@
+Jekyll::Hooks.register :site, :pre_render do |site, payload|
+  FrontMatterTests.process(site, payload)
+end
+
+module FrontMatterTests
+
+  # Checks if key provided in schema is required
+  #
+  #
+  # @param {string} key  - key name
+  # @param {hash} schema - the hash-representation of a schema file
+  #
+  # Returns true or false depending on the success of the check.
+  def self.required?(key, schema)
+    is_required = true
+    is_primary = schema[key]
+    schema['config'] = schema['config'] || { 'optional': [] }
+    is_optional = schema['config']['optional'].include?(key)
+
+    if is_primary && !is_optional
+      is_required
+    elsif (is_primary && is_optional) || (!is_primary && is_optional)
+      !is_required
+    else
+      raise 'The key provided is not in the schema.'
+    end
+  end
+
+  # Processes a collection against a schema
+  #
+  # Opens each file in the collection's expected directory and checks the
+  # file's frontmatter for the expected keys and the expected format of the
+  # values.
+  #
+  # NOTE - As it iterates through files, subdirectories will be ignored
+  #
+  # @param {hash} schema - the hash-representation of a schema file
+  #
+  # Returns true or false depending on the success of the check.
+  def self.processCollection(schema)
+    dir = File.join(schema['config']['path'])
+    passfail = []
+    Dir.open(dir).each do |f|
+      next if File.directory?(File.join(dir, f))
+      file = File.open(File.join(dir, f))
+      next if schema['config']['ignore'].include?(f)
+      data = YAML.load_file(file)
+
+      passfail.push check_keys(data, schema.keys, f)
+      passfail.push check_types(data, schema, File.join(dir, f))
+    end
+    passfail.keep_if { |p| p == false }
+    if passfail.empty?
+      return true
+    else
+      puts "There were #{passfail.count} errors".red
+      return false
+    end
+  end
+
+  # Loads a schema from file.
+  #
+  # @param {string} file - a string containing a filename
+  #
+  # Returns a hash loaded from the YAML doc or exits 1 if no schema file
+  # exists.
+  def self.load_schema(file, config)
+    schema = File.join(Dir.pwd, config['path'], file)
+    if File.exist?(schema)
+      YAML.load_file(schema)
+    else
+      puts "No schema for #{file}"
+      exit 1
+    end
+  end
+
+  # Checks a hash for expected keys
+  #
+  # @param {string} target - the hash under test
+  # @param {string} keys   - an array of keys the data is expected to have, usually loaded from
+  #        a schema file by loadschema()
+  # @param {string} title  - A string representing `data`'s name
+  def self.check_keys(target, keys, title)
+    keys -= ['config']
+    unless target.respond_to?('keys')
+      puts "The file #{title} is missing all frontmatter.".red
+      return false
+    end
+    diff = keys - target.keys
+    if diff.empty?
+      return true
+    else
+      puts "\nThe file #{title} is missing the following keys:".red
+      for k in diff
+        puts "    * #{k}".red
+      end
+      return false
+    end
+  end
+
+  # Validates that the values match expected types
+  #
+  # @param {string} data - the hash under test
+  # @param {hash} schema - the hash-representation of a schema file
+  # @param {string} file - a string containing a filename
+  def self.check_types(data, schema, file)
+    return false unless data.respond_to?('keys')
+    schema.each do |s|
+      key = s[0]
+      value = s[1]
+      type = if value.class == Hash
+               value['type']
+             else
+               value
+             end
+
+      next unless required?(key, schema)
+      if key == 'config'
+        next
+      elsif value.class == Hash
+        dict_key = value.values[1]
+        dict_key_type = value.values[2]
+        if data[key].class == Hash
+          if dict_key_type == 'String' && data[key][dict_key].class == String
+            next
+          else
+            puts "\nThe file #{file} is missing the following keys:\n" \
+                 "    * #{dict_key}".red
+            puts "    * invalid value for '#{dict_key}' in #{file}. " \
+                 "Expected #{dict_key_type} but was #{data[dict_key].class}\n\n"
+            return false
+          end
+        else
+            puts "\nThe file #{file} is missing the following keys:".red
+            puts "    * #{key}".red
+            puts "    * invalid value for '#{key}' in #{file}. " \
+                 "Expected #{type} but was #{data[key].class}\n\n"
+          return false
+        end
+      elsif type == 'Array' && data[key].class == Array
+        next
+      elsif type == 'Boolean' && data[key].is_a?(Boolean)
+        next
+      elsif type == 'String' && data[key].class == String
+        next
+      elsif type == 'Date'
+        next
+      else
+        puts "    * invalid value for '#{key}' in #{file}. " \
+             "Expected #{type} but was #{data[key].class}\n\n"
+        return false
+      end
+    end
+  end
+
+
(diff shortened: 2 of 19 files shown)


mtlynch:
Review status: 0 of 19 files reviewed at latest revision, all discussions resolved.

---

*[_linter/frontmatter-tests/_posts.yml, line 3 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgIZj6AoAixoagUcpG:-LEgIZj6AoAixoagUcpH:b-kr2j44) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_linter/frontmatter-tests/_posts.yml#L3)):*
> ```YAML
> ---
> excerpt: "String"
> header: 
> ```

Trailing whitespace

---

*[_plugins/jekyll_frontmatter_tests.rb, line 7 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgKCux1IAhI3A28AnA:-LEgKCux1IAhI3A28AnB:bpmakxt) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L7)):*
> ```Ruby
> module FrontMatterTests
> 
>   # Checks if key provided in schema is required
> ```

I prefer to avoid passive voice in documentation where possible, because it's harder for most people to read and it often introduces ambiguity by omitting the subject. Suggest just cutting the word "provided".

---

*[_plugins/jekyll_frontmatter_tests.rb, line 13 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgK2f9DFlezMy7s1YS:-LEgK2f9DFlezMy7s1YT:b-7hr2d6) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L13)):*
I feel like this is a bit ambiguous because it doesn't say which result is associated with which return value. Suggest:

>Returns true if frontmatter is valid.

Some people attach something like "Returns true if X, returns false otherwise" but I feel like the second part can be omitted because it's implied.

(ditto for lines 39 and 159)

---

*[_plugins/jekyll_frontmatter_tests.rb, line 25 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgK_2M00nlp_M3xqSW:-LEgK_2M00nlp_M3xqSX:bglx0et) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L25)):*
> ```Ruby
>       !is_required
>     else
>       raise 'The key provided is not in the schema.'
> ```

If this happens, does the error message show which key caused the error? If not, can we include the key name in the error message?

---

*[_plugins/jekyll_frontmatter_tests.rb, line 42 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgKybF6cgn6pml0EoS:-LEgKybF6cgn6pml0EoT:b-cmlq5y) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L42)):*
> ```Ruby
>   def self.processCollection(schema)
>     dir = File.join(schema['config']['path'])
>     passfail = []
> ```

Is there a more descriptive name we could use for `passfail`? Just from the variable name, I have a hard time understanding what it would store.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 61 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLAtaA1uqiT1dT1jk:-LEgLAtaA1uqiT1dT1jl:br4qnng) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L61)):*
> ```Ruby
>   end
> 
>   # Loads a schema from file.
> ```

This has a trailing period but the rest of the one-line summaries do not. Suggest using a trailing period on all of them for consistency.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 63 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLKIC4dO0URpjiN3H:-LEgLKIC4dO0URpjiN3I:b-5nsc55) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L63)):*
> ```Ruby
>   # Loads a schema from file.
>   #
>   # @param {string} file - a string containing a filename
> ```

Suggest `file` -> `schema_filename`

---

*[_plugins/jekyll_frontmatter_tests.rb, line 64 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLOYQ7e3urn896NVt:-LEgLOYQ7e3urn896NVu:bbvop9k) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L64)):*
> ```Ruby
>   #
>   # @param {string} file - a string containing a filename
>   #
> ```

Can we add documentation for `config`?

---

*[_plugins/jekyll_frontmatter_tests.rb, line 80 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgIhYoEtm8Z95JK41J:-LEgIhYoEtm8Z95JK41K:b-2191dn) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L80)):*
> ```Ruby
>   #
>   # @param {string} target - the hash under test
>   # @param {string} keys   - an array of keys the data is expected to have, usually loaded from
> ```

The rest of this file seems to break lines at 80 columns. Suggest breaking at 80 columns here for consistency.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 82 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLd-l1gIC6e0P7tD2:-LEgLd-l1gIC6e0P7tD3:bhu0cdl) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L82)):*
> ```Ruby
>   # @param {string} keys   - an array of keys the data is expected to have, usually loaded from
>   #        a schema file by loadschema()
>   # @param {string} title  - A string representing `data`'s name
> ```

Can we elaborate on this? I'm not sure what this means.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 83 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLuHl4VRt7j0PXrQN:-LEgLuHl4VRt7j0PXrQO:b-mbrnaj) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L83)):*
> ```Ruby
>   #        a schema file by loadschema()
>   # @param {string} title  - A string representing `data`'s name
>   def self.check_keys(target, keys, title)
> ```

Can we document the return value as well?

(Ditto for `check_types`)

---

*[_plugins/jekyll_frontmatter_tests.rb, line 101 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgLkwn05O7v6e4JzuB:-LEgLkwn05O7v6e4JzuC:b27uy3l) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L101)):*
> ```Ruby
>   end
> 
>   # Validates that the values match expected types
> ```

Can we elaborate a bit here? This is a bit vague.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 157 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgJgH06LT-wMXcAUs0:-LEgJgH06LT-wMXcAUs1:b-s82aqq) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L157)):*
Can we rephrase this in terms of what this actually tests for? Like:

>Validates  frontmatter for a post to ensure it has X, Y, and Z.

---

*[_plugins/jekyll_frontmatter_tests.rb, line 159 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgM6Md67a9Itupsgb4:-LEgM6Md67a9Itupsgb5:blzn9su) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L159)):*
> ```Ruby
>   # Initial function that starts the frontmatter tests
>   #
>   # Returns true or false depending on the success of the checks.
> ```

Can we document the parameters?

---

*[_plugins/jekyll_frontmatter_tests.rb, line 160 at r1](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-LEgIpVx-E48W97UBXUw:-LEgIpVx-E48W97UBXUx:bxeyrl8) ([raw file](https://github.com/mtlynch/mtlynch.io/blob/8d4e1ebd5a9e10a5304f8d04af564f94910bcab3/_plugins/jekyll_frontmatter_tests.rb#L160)):*
> ```Ruby
>   #
>   # Returns true or false depending on the success of the checks.
>   def self.process(site, payload)
> ```

(I don't know Ruby so apologies for the Ruby ignorance)

Is it possible to make this function  the first in the module? I like top-down ordering where possible.

Also, I'm not sure how this works in Ruby but are any of these functions callable from outside this file? Is it possible to make everything except `process` "private` given that it's the only public interface to this module?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mtlynch/mtlynch.io/266)*

mtlynch:
This is great! I added some notes to help me understand the checker implementation.

mtlynch:
:lgtm: thanks for the fixes!

---

Review status: 0 of 19 files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/mtlynch/mtlynch.io/266#-:-LF8JQuqDEU0oFQsoyqq:b-cun82c)*

PR #275, opened 2018-07-10
BUG: Fixing build dependency failures

### LINKED ISSUE #274: Fix build break
The build is currently broken:

https://travis-ci.org/mtlynch/mtlynch.io/builds/401815972

```text
Installing nokogiri 1.8.4 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
    current directory: /var/lib/gems/2.3.0/gems/nokogiri-1.8.4/ext/nokogiri
/usr/bin/ruby2.3 -r ./siteconf20180709-71-155n7f6.rb extconf.rb
checking if the C compiler accepts ... yes
Building nokogiri using system libraries.
pkg-config could not be used to find libxml-2.0
Please install either `pkg-config` or the pkg-config gem per
    gem install pkg-config -v "~> 1.1"
pkg-config could not be used to find libxslt
Please install either `pkg-config` or the pkg-config gem per
    gem install pkg-config -v "~> 1.1"
pkg-config could not be used to find libexslt
Please install either `pkg-config` or the pkg-config gem per
    gem install pkg-config -v "~> 1.1"
ERROR: cannot discover where libxml2 is located on your system. please make sure
`pkg-config` is installed.
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.
Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/usr/bin/$(RUBY_BASE_NAME)2.3
	--help
	--clean
	--use-system-libraries=true
	--with-zlib-dir
	--without-zlib-dir
	--with-zlib-include
	--without-zlib-include=${zlib-dir}/include
	--with-zlib-lib
	--without-zlib-lib=${zlib-dir}/lib
	--with-xml2-dir
	--without-xml2-dir
	--with-xml2-include
	--without-xml2-include=${xml2-dir}/include
	--with-xml2-lib
	--without-xml2-lib=${xml2-dir}/lib
	--with-libxml-2.0-config
	--without-libxml-2.0-config
	--with-pkg-config
	--without-pkg-config
	--with-xslt-dir
	--without-xslt-dir
	--with-xslt-include
	--without-xslt-include=${xslt-dir}/include
	--with-xslt-lib
	--without-xslt-lib=${xslt-dir}/lib
	--with-libxslt-config
	--without-libxslt-config
	--with-exslt-dir
	--without-exslt-dir
	--with-exslt-include
	--without-exslt-include=${exslt-dir}/include
	--with-exslt-lib
	--without-exslt-lib=${exslt-dir}/lib
	--with-libexslt-config
	--without-libexslt-config
To see why this extension failed to compile, please check the mkmf.log which can
be found here:
  /var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/nokogiri-1.8.4/mkmf.log
extconf failed, exit code 1
Gem files will remain installed in /var/lib/gems/2.3.0/gems/nokogiri-1.8.4 for
inspection.
Results logged to
/var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/nokogiri-1.8.4/gem_make.out
An error occurred while installing nokogiri (1.8.4), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.8.4'` succeeds before bundling.
In Gemfile:
  minimal-mistakes-jekyll was resolved to 4.9.0, which depends on
    jemoji was resolved to 0.10.0, which depends on
      html-pipeline was resolved to 2.8.3, which depends on
        nokogiri
The command "_tests/docker_build" exited with 5.
```

Not sure what's happening as nothing changed since the last build.

I messed up and lost the log for the [last successful build](https://travis-ci.org/mtlynch/mtlynch.io/builds/393097074) in Travis by restarting it, but it built successfully on 2018-06-06 and nothing in the source changed in the meantime.

I can update [the Docker image](https://hub.docker.com/r/mtlynch/jekyll/) if that would be helpful. I started to, then realized it would probably make it easier to debug the issue if we didn't alter anything, so I left the old image up.

shredtechular:
This PR fixes #274 

## Issue
Based on the build log file/link provided, we can see that the build is failing on the install of the `nokogiri` dependency, which I know is specifically used with the `html-proofer` gem.  Looking back at previous builds that were successful, we can see that the nokogiri was at `1.8.1` or `1.8.2` and the current build failure is at `1.8.4`.  With `nokogiri` having recently released a newer version, bundler is automatically determining to use the newer version instead of the tried and true version with the current setup.

## Fix
The fix included with this PR is to include `Gemfile.lock` in the codebase to reduce these types of dependency issues.  Best practice, from most of the guidance I've read, seems to now recommend including the lock file in applications, but not recommended for actual Gem development. Source: <https://jekyllrb.com/tutorials/using-jekyll-with-bundler/> and <https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/>.

Checking in the lock file resolved the build tests on my end, however, I did get some intermittent and unrelated external link failures that looks like are being addressed in #272, but will submit this and check it to make sure everything works properly.

With this solution, running `_tests/docker_build` locally will also change `Gemfile.lock` since we are altering the `Gemfile` in the build commands, so also wanted to point that out.  I think this is minor, but will need to be careful about running that locally and not checking in updated `Gemfile` and `Gemfile.lock` files unintentionally.

**Please note** - At some point, I expect we may be forced to use a newer version of `nokogiri,` especially if we want to use a newer version of `html-proofer`.  I'm of the belief that we cross that if/when this happens?  I do believe we can tweak `Dockerfile` to work with the newer version of `nokogiri`, but I'd recommend holding off on that until we deem it necessary as maybe the next version of `nokogiri` changes the installation steps again.  So, I don't think we need to worry about that unless we have a need to upgrade.  If you feel differently, we can work on updating the `Dockerfile` as well to work with the newer version of nokogiri, but at least by adding the `Gemfile.lock` builds can continue as necessary.


mtlynch:
Ah, I like having the `Gemfile.lock` in there. I'm used to npm's `package-lock.json` and I forgot there was an equivalent for gems.

PR #286, opened 2018-07-18
Feature: Adding advertisement include functionality

### LINKED ISSUE #285: Add an include for Zestful ads
I'd like to be able to have native ads for my current project, [Zestful](https://zestfuldata.com). I'd like it to look kind of like the UI element for newsletter subscription:

image

I'd like the text to be customizable per-include (kind of like image includes), but as placeholder, we can start with the text:

---
**Want to parse ingredients without all this work?**

I went through all of these steps so you don't have to. Check out [Zestful](https://zestfuldata.com), my managed service for ingredient parsing.

---

I'd like the logo to appear as well. Here it is with the standard background:

* [Logo (with background)](https://drive.google.com/open?id=1uLgz0JqfdvAd0IzFDfx71OIKYhESNRLs)

If the standard logo doesn't look good on the site, feel free to play with the colors starting from this version:

* [Logo (transparent background)](https://drive.google.com/open?id=1fjNPYzUzT5J5ez1edJd6Zo5R0a1CVVdg)

shredtechular:
This PR fixes #285 

This PR is an initial cut at adding a generic advertisement include functionality that should be pretty flexible and customizable so that it doesn't always have to apply for the Zestful.  The styling is all pretty static (and may need additional styling depending on the elements used in future advertisements...) but all of the content in the advertisement is markdown based, so should be easier to edit and add new elements like lists, multiple images, different headers, etc. if any other types of information were needed to be displayed.

## More detailed explanation of how it works
This solution was coded to minimize the amount of effort needed to include an advertisement in any given post as well as to make it as flexible as possible to add/edit content, images, etc.  In order to do that, I chose for the actual advertisement content to be in markdown form and created a new collection called sections (titled "sections" so, in the future, other types of content blocks or includes could be added in this way as well...).  In the markdown file, it has 1 require piece of yml data:
```
---
type: ad
---
```
The `type:` key will always be `ad` for advertisements (again, if other types of content blocks are desired, we can add new types).

And of course, then the content is in markdown format.

Then the final step would be to add the specific content block into the given post, like so:
`{% include ads.html title="title_of_markdown_file" %}`
so for example, to include the canned "Zestful" advertisement that is located at `_sections/ads/zestful.md` into a blog post, you can simply do:
`{% include ads.html title="zestful" %}`
The `title=` parameter in the include statement matches the filename in the sections collection.

Which would look like the below screenshot:
screen shot 2018-07-19 at 9 13 45 pm

### Testing
If you would like to see this in action before merging, you can pull down my feature branch and simply add `{% include ads.html title="zestful" %}` into any post that you are thinking about displaying the ad to see how it looks and works.  I wasn't sure where you were thinking about putting it though so purely building this branch would not render the advertisement anywhere since I didn't commit the include into any posts.

### Other Important Notes
- Unfortunately, the responsive-image tag isn't going to work inside the include.  Apparently this only works in layouts instead of includes and just can't find a way around that...
- ~~I made the **whole entire advertisement** a big link instead of just the "Zestful" word being a link as the issue may have desired?  That can be changed very easily in the markdown however by removing the `link:` yml frontmatter key and then making the word "Zestful" linkable.  I thought most advertisements make the whole section clickable though, so made it like that instead?~~

### Alternative solution
If you'd rather not deal with adding separate markdown files for different advertisements, then we can do something more like:
`{% include ads.html image="zestful-logo.png" heading="Want to parse ingredients without all this work?" subtext="I went through all of these steps so you don't have to. Check out Zestful, my managed service for ingredient parsing." %}`
But I felt it was a little more limiting and tedious?  But if that is easier, I can easily alter the solution to more of that form.


mtlynch:
Cool, I like the code structure for this.

I'd like the ad itself to be a little less intrusive/jarring. Could we do something a little slimmer like this?

zestful-mockup

I'd like to avoid what people generally think of as "banner ad" and I feel like having the whole thing be a giant button will make people think banner ad. Can we make only the logo itself and the word "Zestful" links? And can we try a more subtle banner background (like a lighter blue or gray) to clue the reader that only certain parts are links? I'd still want the normal blue background for the logo itself.

shredtechular:
That makes sense.  I have slimmed down the ad in 30d20fa8f3f5f4e790436f7beb81515996d5b86d and used the gray background for the entire ad.  Then updated the Zestful ad with the blue background.  I also removed the whole ad to be clickable and just made the logo and the word "Zestful" links which are all now controlled in the markdown.

Below are a couple of updated screenshots.

Desktop:
<kbd>
screen shot 2018-07-19 at 9 13 45 pm
</kbd>

Mobile:
<kbd>
screen shot 2018-07-19 at 9 14 08 pm
</kbd>

**I updated the original PR with all of this information as well**

mtlynch:
This looks good, but I think some files might be missing from the PR. It refers to `images/zestful-logo.png'` but the PR only has `images/Zestful_logo_no_background.png`, which isn't the same image as what's in the screenshots.

mtlynch:
I tested this on Chrome and only the bottom half of the image is a clickable link. It looks like the `h3` is accidentally covering it up:

image

Would it be possible to get it so that the full image is clickable?

shredtechular:
Absolutely, that is very doable and I should've caught that.  Very nice to find that!  It's a simple CSS fix to ensure the `h3` doesn't overlap the logo which is now complete in 08d1ac0.

mtlynch:
Awesome, looks great! Thanks!

PR #304, opened 2018-08-11
Feature: Updating "jekyll-admin" and adding livereload functionality

### LINKED ISSUE #276: Configure for --livereload
It looks like `--livereload` is [now supported](https://github.com/jekyll/jekyll-admin/pull/468) in jekyll-admin. Can we update packages / change `script/_serve_dev.sh` to support this?

shredtechular:
Fixes #276 

The code changes with this update enable the ["livereload" functionality for Jekyll](https://jekyllrb.com/news/2018/01/02/jekyll-3-7-0-released/).  In order to accomplish this, we need to make sure the "jekyll-admin" gem uses the latest 0.8.1 version which is changed in the `Gemfile.lock`.  In order to take advantage of this feature, a `bundle install` will be necessary to ensure the new version of that gem is used.

Then it was just a matter of adding the `--livereload` flag to the development script.  So, once the site is built and the browser(s) are loaded and navigated to the local site, then any subsequent content, layout, css changes that are made will reload the browser(s) automatically.  Very cool!  Will make cross-browser testing a bit easier!

In my testing of this feature, I tested some different changes to ensure the browser(s) were automatically reloaded and saw the correct changes:
- [X] Content changes within `_posts` or `_pages` or `index.html` worked properly.
- [X] Adding new images via the image include tag automatically reloaded the browser(s).
- [X] Adding new file includes inside content worked properly reloading automatically.
- [X] Changing SASS/CSS also reloaded the browser(s) automatically
- [ ] Changing the javascript is the only scenario that's a little funky.  This is obviously because the js is built outside the Jekyll scope using the `npm run build:js` command.  It does reload automatically if I run `npm run watch:js` in a separate terminal, which is awesome, but Jekyll sees 2 changes (one for initial js change and one for the compiled js, so it rebuilds and reloads twice.  I tried tweaking the `exclude` config values to exclude the js partials from being "watched" by Jekyll, but its still building/reloading twice.  I can keep looking at this if you'd like me to spend more time on it, but I think this is a minor edge case and probably isn't terribly important.  I think the key is that it reloads well based on content, layout, and CSS changes as those are where the majority of the changes take place.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -56,7 +56,7 @@ GEM
       pathutil (~> 0.9)
       rouge (>= 1.7, < 4)
       safe_yaml (~> 1.0)
-    jekyll-admin (0.8.0)
+    jekyll-admin (0.8.1)
       addressable (~> 2.4)
       jekyll (~> 3.3)
       sinatra (~> 1.4)
--- a/script/_serve_dev.sh
+++ b/script/_serve_dev.sh
@@ -9,5 +9,6 @@ npm run build:js
 # Start jekyll server
 bundle exec jekyll serve \
   --config _config.yml,_config_dev.yml \
+  --livereload \
   --watch \
   --incremental


mtlynch:
This is awesome! Way better than what I was doing before, thanks!

>I can keep looking at this if you'd like me to spend more time on it, but I think this is a minor edge case and probably isn't terribly important. I think the key is that it reloads well based on content, layout, and CSS changes as those are where the majority of the changes take place.

Yeah, you made the right call here. I rarely edit JS on mtlynch.io. I actually don't think I've ever done it. It's sufficient to just live reload on content changes, so the CSS and layout too is a bonus. JS isn't necessary.

PR #305, opened 2018-08-12
Feature: Removing dates from image paths

### LINKED ISSUE #278: Remove dates from images paths
I like how `_files` subdirectories are named by permalink rather than date+permalink. Can we do the same thing for the `images` directory so that the subfolders are just named by permalink?

shredtechular:
Fixes #278.

This fix removes the dates from the graphics stored in the `images` directory.  In order to accomplish this, we obviously needed to ensure the images didn't become broken since they were expecting the path to have the post's date in the directory name.  So, the only code change here was to simplify the image include to simply use the `page.url` variable instead of using `page.path` and having to manipulate it.  So, this is a simpler solution in all.

There was one image that also had the date prefix in the content as well, so I did need to fix that path as well.

Aside from those 2 files, the rest of the 184 files that were changed were simply renaming the image post directories to remove the date from them.


mtlynch:
Awesome, thanks!

PR #312, opened 2018-08-16
Styling: Reducing bottom margin on figures

### LINKED ISSUE #309: img-border creates empty space at bottom of image
I just realized that when `fig_caption` is set, the `img-border` class leaves a gap of empty space at the bottom of an image between the image and the border:

image

Would it be possible to close this gap and eliminate the empty space?

shredtechular:
Fixes #309.

This styling update removes the bottom margin that is added to images within `<figure>` tags which are used for captions.

You can see the below/after screenshots below:
### Before
screen shot 2018-08-15 at 8 22 46 pm

### After
screen shot 2018-08-15 at 8 23 21 pm

### DIFF
--- a/_sass/child-theme/base/_base.scss
+++ b/_sass/child-theme/base/_base.scss
@@ -8,6 +8,10 @@ figure {
   display: table;
   margin: 0em auto 1.3em;
   font-size: 0.9em;
+
+  img {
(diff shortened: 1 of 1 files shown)


mtlynch:
Nice, thanks!

PR #324, opened 2018-09-15
Styling: Adjusting lists next to floated images

### LINKED ISSUE #323: [Book Reports] Fix layout issues with images
I'm adding a book reports feature to the site (see [book reports branch](https://github.com/mtlynch/mtlynch.io/tree/book-reports)).

There are lots of places we can optimize later to make this feature work better (eliminate redundant images, redundant text, auto-index), but for now, I'm fine with quick 'n dirty hacks.

There are two spots where it's a bit too dirty:

image

image

The bullet points are crowding the image too much. Could we space them out more?

I'm good with a fix that addresses it generally site-wide or a special-case fix for the book reports. I'm also open to other ways of presenting the information in the bullet points. If there's a way that's more CSS heavy and prettier, I'm down for that, but mainly just aiming for whatever lets us publish the soonest, then we'll revisit and smooth it out.

You might notice that the horizontal spacer is covered up. That's not a big issue for now because there will actually be more text in the blurb, so that shouldn't occur in practice.

shredtechular:
Fixes #323 

This PR fixes the spacing issue identified in #323 where lists next to floated images have the bullets not spaced nicely.  This solution adds a `overflow:hidden` rule to lists which forces a new block formatting context for the list's contents and thus positioning said block next to the floated image.

This fix is the most quick and dirty solution that I can think of that:
1) won't impact other lists in the site
2) minimizes the use of a special class that could hamper add additional complexity for entering lists in markdown
3) achieves the nicer spacing look that lists across the site currently use -- consistency

The only issues this might present are:
1) if the list next to the floated image is very long, the list won't "float" around the image, which may or may not be desired, but does defeat the purpose of "floated" images if the list doesn't "float" around it technically I suppose...
2) I think very old versions of IE this wouldn't work on... like IE6/7...  tested on Edge & IE11 and looks good though.

New spacing with new CSS rule applied:
screen shot 2018-09-14 at 9 29 31 pm

Same spacing on other lists even with the new CSS rule applied:
screen shot 2018-09-14 at 9 29 51 pm

### DIFF
--- a/_sass/child-theme/base/_base.scss
+++ b/_sass/child-theme/base/_base.scss
@@ -23,3 +23,7 @@ img {
 a {
   text-decoration: none;
 }
+
+ul, ol {
+  overflow: hidden;
+}


mtlynch:
Nice, thanks!

>if the list next to the floated image is very long, the list won't "float" around the image, which may or may not be desired, but does defeat the purpose of "floated" images if the list doesn't "float" around it technically I suppose...

Yeah, that's okay. I can't think of a case where I'd want a list to wrap like that.

>I think very old versions of IE this wouldn't work on... like IE6/7... tested on Edge & IE11 and looks good though.

Yeah, that's fine. We've actually never discussed desired browser compatibility. I checked my visitor stats and Edge + IE only make up ~2% of my total traffic, so it's not even worth compatibility testing for those.

I've written up an [explicit strategy for compatibility testing](https://github.com/mtlynch/mtlynch.io#compatibility-targets). I excluded Edge/IE from the compatibility targets (though, as a Microsoft loyalist, I feel guilty).

shredtechular:
> I've written up an explicit strategy for compatibility testing. I excluded Edge/IE from the compatibility targets (though, as a Microsoft loyalist, I feel guilty).

Awesome idea and very helpful!

PR #339, opened 2018-10-14
Feature: Refactoring book reports index page

### LINKED ISSUE #330: [Book Reports] Generate index page automatically
Right now, the book report index page is generated manually with a lot of hacks, copy/pasted code, and duplicate images. I'd like to eliminate the redundancy and let Jekyll generate the index page automatically based on the available book report Markdown files available.

shredtechular:
Fixes #330 

# One IMPORTANT note to start:
I was receiving htmlproofer build errors on a few different links in the `_posts/2018-08-20-resurrecting-3.md`.  It appears as though the coverage.py documentation has changed its documentation/URLs.  So, I did my best to update these URLs in that post, but I'm not sure if I got these URLs fixed as desired?  If I should remove the commit that has those updates from this PR, I certainly can, or can change them to something more preferable...

----

This PR addresses #330 to remove much of the manual work behind generating the [book-reports index page](https://mtlynch.io/book-reports/).  

As discussed in the issue, the parent theme has incorporated a new layout that will display [collections dynamically on an index page](https://mmistakes.github.io/minimal-mistakes/docs/layouts/#layout-collection).  So, in order to accomplish this feature, the first task was to upgrade the parent theme to the most recent version (4.13.0).  Doing that, resulted in a few minor styling changes, like font sizes and spacing.  I highlighted a couple of the bigger ones that struck out to me below, if there are any others that are not desired, we can attempt to override the parent theme styling:

1. All templates/pages should have the same gutter widths as shown in screen shots below:
### Current Site
gutters-before

### With Updated theme
gutters-after

2. Blog index page links now match the same color as other links throughout the site for consistency as seen below:
### Current Site
screencapture-mtlynch-io-2018-10-13-20_09_01

### With Updated theme
screencapture-localhost-4000-2018-10-13-20_27_36

Once the theme had been satisfactorily updated, the next step was to ensure the book reports excerpts included all of the required information.  In order to most efficiently do that, I added an "excerpt separator" of `<!---more--->` into each of the book report pages where the content is located.  As the excerpt separator is a comment, this will not affect the layout of the actual book report pages.  I also updated the index page layouts & added some additional javascript to get some of the information that wasn't in the book report markdown files to meet the required information.  

The last step was ensuring the sort order was correct.  I added a `read_date` frontmatter field where that date can be stored for each book report and also tweaked the parent theme's sorting capability to be able to sort on custom frontmatter fields (this is something I plan to submit a PR for into the parent theme to try and get accepted...and if it does get accepted we can remove that file from this child theme.


_includes/archive-single.html
```diff
@@ -36,6 +36,23 @@ <h2 class="archive__item-title" itemprop="headline">
     {% if post.read_time %}
       <p class="page__meta"><i class="far fa-clock" aria-hidden="true"></i> {% include read-time.html %}</p>
     {% endif %}
-    {% if post.excerpt %}<p class="archive__item-excerpt" itemprop="description">{{ post.excerpt | markdownify | strip_html | truncate: 160 }}</p>{% endif %}
+    {% if post.collection == "book_reports" %}
+      <hr />
+      
+      {% comment %}
+        Split the post's excerpt into two parts using the unordered list
```
mtlynch: Line is breaking a little too early. Line limit is <= 80 characters, so keeping " element" on the same line makes it exactly 80 characters.

assets/js/_main.js
```diff
@@ -69,4 +69,12 @@ $(document).ready(function(){
     midClick: true // allow opening popup on middle mouse click. Always set it to true if you don't provide alternative source.
   });
 
+  // Change links on Book-reports Index page from amazon link to post url
```
mtlynch: Can we elaborate a bit on this comment? I don't understand what this means.

mtlynch:
>So, I did my best to update these URLs in that post, but I'm not sure if I got these URLs fixed as desired?

Aw man. I always think I have Travis set up to run builds daily and notify me when it breaks but there's always something wrong. It started failing two weeks ago and never notified me:

image

I applied a slightly different fix for this in #340. Can you revert the change to that file and rebase on master?

>The last step was ensuring the sort order was correct. I added a read_date frontmatter field where that date can be stored for each book report and also tweaked the parent theme's sorting capability to be able to sort on custom frontmatter fields (this is something I plan to submit a PR for into the parent theme to try and get accepted...and if it does get accepted we can remove that file from this child theme.

Awesome, good idea!

_includes/archive-single.html
shredtechular: Hmm, weird it shows as 81 characters in my editor with "element" on the first line?  I updated it so it is on the first line now though in aacec1cfb4e333c43d9cfb15464fc84077068c10.

assets/js/_main.js
shredtechular: Yeah, that wasn't very good/descriptive reading that back now... good catch.

Basically what this is doing is changing the links around the images of the books on the book report **index page** so that they aren't pointing to the external (currently both books have Amazon links...) link, but instead linking to the actual book report page.  On the book report page, the link will still point to the external URL (Amazon currently...).  So, this only changes the link for navigation purposes on the index page.  This is how the current book report index page works as well.  I know you aren't married to that necessarily, but it seemed to make sense to me so liked to keep that in there?  Can certainly take it out completely if you are okay with the links around those images linking to Amazon even on the book report index page...

Does that make sense?  Screen capture below to help illustrate as well the links the js is changing dynamically:
screen shot 2018-10-13 at 10 57 11 pm

I updated the comment to say "Loop through the image links on the book-reports index page and change them from pointing to external link to pointing to the individual book report page associated with that image." in aacec1cfb4e333c43d9cfb15464fc84077068c10.

Is that better/more descriptive?  Any thoughts on how better to describe it?

shredtechular:
> I applied a slightly different fix for this in #340. Can you revert the change to that file and rebase on master?

Absolutely, no problem.  That is complete now and I updated the PR to remove the note about those links as well.

PR #355, opened 2019-01-23
Refactoring files include to match new liquid style guide

mtlynch:


_includes/files.html
```diff
@@ -32,32 +32,50 @@
...
 
-  {% if subdir == page.slug %}
+      /files/sia-nextcloud/docker-compose
+
+    becomes:
+
+      /files/sia-nextcloud/docker-compose.yml
+  {% endcomment %}
+  {% if include.language == 'yml' or include.language == 'yaml' %}
+    {% assign file_relative_url = file_relative_url | append: '.' | append: include.language %}
```
shredtechular: I can update this in the subsequent task for this code, but upon review with fresh eyes, this needs to change and I want to make sure I make a not of it.  I think the last append here should be hardcoded to "yml" because if the include.language is set to "yaml" then the embed of the file fails.

So this line should more simply become:
`{% assign file_relative_url = file_relative_url | append: '.yml' %}`

mtlynch: Ah, good catch. Fixed in 93bf0d2801041c8706e865d176f9acca3a6d9512.

PR #357, opened 2019-01-27
Bug: File embed race condition refactor

### LINKED ISSUE #310: Liquid exception finding multiple files even though there's only one
I'm seeing what seems like it might be a race condition with the liquid exception handling for file includes:

https://gist.github.com/mtlynch/2de09ceeaf1c55193f33c81dcfe62e55

Every few saves, it gets into a state where it claims that multiple includes have the same title, but the titles seem to be configured correctly.

The errors are also accompanied by lines like this:

```
/usr/local/lib/ruby/gems/2.4.0/gems/jekyll-3.8.2/lib/jekyll/reader.rb:42: warning: conflicting chdir during another chdir block
```

which makes me think there's maybe a race condition (although I've been seeing Jekyll produce that warning, seemingly at random, for many months now).

This is on Ubuntu 16.

```
$ ruby --version
ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-linux]
```

shredtechular:
Fixes #310.

This PR also supercedes #316.

I thought it would be cleaner to submit a distinct PR for the refactoring of the file embed race condition remediation discussed in #316 rather than adding to that existing PR.  If it is desired to keep this all in #316, I can add this to that PR.

This update includes the changes in #316 to remediate the race condition on the file embed include so that false positive errors are not thrown.  This is a "workaround" to the race condition, however, it also ensures duplicate embeds don't happen either so there is progress with this update even if the race condition did not exist.

## What this workaround WILL do
This workaround will:
- [X] Ensure there are no false positives for duplicate file include titles by creating an array of file paths and checking to ensure the same file path is not "counted" as a duplicate file title
- [X] Provide the intended and original functionality for pro-actively finding duplicate and/or incorrect file include titles

## What this workaround DOESN'T do
This workaround doesn't:
- [ ] Address the `conflicting chdir during another chdir block` build warning.  This appears to be at the root of the issue, but again, only occurs when using `Jekyll-admin` for editing content.  So, I believe, the race condition is still present, but doesn't affect how the site is built either for production or when using `Jekll-admin` and with this update, at least false positives are not thrown.

### DIFF
--- a/_includes/files.html
+++ b/_includes/files.html
@@ -29,6 +29,12 @@
 {% comment %} Initialize file count. {% endcomment %}
 {% assign file_count = 0 %}
 
+{% comment %} 
+  Initialize an empty string to store unique file paths that match the correct
+  file path and title.
+{% endcomment %}
+{% assign found_files = "" %}
+
 {% for file in files %}
 
   {% comment %}
@@ -70,6 +76,13 @@
     {% assign file_relative_url = file_relative_url | append: '.yml' %}
   {% endif %}
 
+  {% comment %}
+    Check to see if any files have been stored (or found) yet.  If string is
+    empty, set it to the current file's path.
+  {% endcomment %}
+  {% if found_files == "" %}
+    {% assign found_files = file_relative_url %}
+
 ```{{ lang_highlight }}
 {{ file.content -}}
 ```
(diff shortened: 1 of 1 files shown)


shredtechular:
I keep getting the below error trying to do anything in Reviewable: 
screen shot 2019-02-02 at 9 06 59 am



So, just adding comment directly in GitHub.  Added the expected condition in the if-else statement in 12a68ee1e4546baab89d627a8aaf80f92dbecc32.

PR #361, opened 2019-02-06
Fixing Twitter cards for book reports

### LINKED ISSUE #359: Book Reports: Generate a better twitter card
The current twitter card for a book report is kind of klunky:

image

I'd like it to use this same card format, but the image should be the book cover and the text should be the start of the blurb.

image

shredtechular:
Fixes #359 

This update is my approach to fixing the twitter cards for the book reports collection to meet what is requested in #359.  This approach:

- Moves the book cover image to the theme's pre-defined frontmatter fields for twitter and open graph images.  So, instead of defining the book cover image in the markdown, defining it in the theme's pre-defined frontmatter fields for open graph and social media sharing resolves the problem of no image/the site's default image displaying in Twitter cards.
- Because the move to using frontmatter fields for the book cover, we also need to add logic to the html template to render the image in the necessary part of the book report page as well.
- Finally, to ensure the book's rating scores and Amazon links are not included in the text of the Twitter card, we can similarly move this to using yaml frontmatter as well.

## Other Notes
- I also added a book report schema for validating `header.teaser` frontmatter values as that should probably always be defined even if null.  We may want to consider validating for the `header.og_image` , `report_intro.score`, `report_intro.link_title`, and `report_intro.link_url` as well.  But I think it's probably best to ensure the approach is amenable from an overall perspective.
- It seems like currently, the alt tag for the book cover image is always the value of the `title` yaml value so I used that in the template.

## Screenshot of locally validating twitter card

screen shot 2019-02-05 at 9 15 06 pm

### DIFF
--- a/_book_reports/fooled-by-randomness.md
+++ b/_book_reports/fooled-by-randomness.md
@@ -5,12 +5,14 @@ author_profile: true
 sidebar:
   nav: main
 read_date: 2018-09-04
+header:
+  teaser: /images/book-reports/fooled-by-randomness/fooled-by-randomness.jpg
+  og_image: ""
+report_intro: 
+  score: 5
+  link_title: Amazon link
+  link_url: https://amzn.to/2x2sPT3
 ---
-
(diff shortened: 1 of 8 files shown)


shredtechular:
Closing this for now as seeing an issue I didn't see previously.  This needs a little more work still...

PR #364, opened 2019-02-07
Fixing Twitter cards for book reports - Take 2

### LINKED ISSUE #359: Book Reports: Generate a better twitter card

shredtechular:
Fixes #359 

This update is my approach to fixing the twitter cards for the book reports collection to meet what is requested in #359.  This approach:

- Moves the book cover image to the theme's pre-defined frontmatter fields for twitter and open graph images.  So, instead of defining the book cover image in the markdown, defining it in the theme's pre-defined frontmatter fields for open graph and social media sharing resolves the problem of no image/the site's default image displaying in Twitter cards.
- Because we move these fields into the frontmatter, it changes how we get the information for the book report index & individual pages for display.  Since the layouts are so similar, I think it would be best to abstract that to it's own include to reduce redundancy.  So we have a new include called `includes/book-report-introduction.html` that contains that template & logic.

## Other Notes
- I also added a book report schema for validating `header.teaser` frontmatter values as that should probably always be defined even if null.  We may want to consider validating for the `header.og_image` , `report_intro.score`, `report_intro.link_title`, and `report_intro.link_url` as well.  But I think it's probably best to ensure the approach is amenable from an overall perspective.
- It seems like currently, the alt tag for the book cover image is always the value of the `title` yaml value so I used that in the template.

## Screenshot of locally validating twitter card

screen shot 2019-02-05 at 9 15 06 pm

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/mtlynch.io/364)

### DIFF
--- a/_book_reports/fooled-by-randomness.md
+++ b/_book_reports/fooled-by-randomness.md
@@ -5,12 +5,14 @@ author_profile: true
 sidebar:
   nav: main
 read_date: 2018-09-04
+header:
+  teaser: /images/book-reports/fooled-by-randomness/fooled-by-randomness.jpg
+  og_image: ""
+report_intro: 
+  score: 5
+  link_title: Amazon link
+  link_url: https://amzn.to/2x2sPT3
 ---
-
-{% include image.html file="fooled-by-randomness.jpg" alt="Fooled by Randomness by Nassim Nicholas Taleb"  link_url="https://amzn.to/2x2sPT3" max_width="250px" class="align-left" %}
-
-* My rating: **5** / 10
-* [Amazon link](https://amzn.to/2x2sPT3)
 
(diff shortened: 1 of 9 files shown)


mtlynch:
Not sure if this pings you with a notification, but you'll want to accept this simple PR to make the build pass: https://github.com/shredtechular/mtlynch.io/pull/1

shredtechular:
Got it, thanks @mtlynch!  I'll also plan to rebase master with the next round of feedback as well to ensure no conflicts.

mtlynch:
>for some reason I'm having trouble commenting on all of the initial comments as a couple of them are no longer displaying...maybe the commit automatically addresses some comments?

I marked some of them as "resolved" based on previous discussion, so it might be because of that. Hard to say, as Reviewable doesn't have the most intuitive UX sometimes.

PR #450, opened 2019-10-05
Fix typo

vzhou842:
Not sure if this was intended? Fixed typo though


mtlynch:
Oh, good catch. Thanks @vzhou842!

PR #546, opened 2020-03-14
Switching from htmlproofer to htmltest

mtlynch:
And fixing a bunch of errors that htmlproofer missed.


mtlynch:
This is blocked on https://github.com/wjdp/htmltest/issues/12

PR #690, opened 2020-11-27
Add vulnerable cpp

mtlynch:


parser/vuln.cpp
```diff
@@ -0,0 +1,34 @@
...
+	size_t nameLen;
+	char* newName;
+	size_t newNameLen;
+};
+
+
+HatManager::HatManager() {}
+
+void HatManager::Rename() {
+    strncpy(name, newName, newNameLen);
```
mtlynch, line 27: There's a buffer overflow here, since we never verify that we allocated enough memory in `name` to fit `newNameLen` characters.

TheWaySnrubThinks: In **my** code? Impossible!

The constructor calls `PurchaseHats`, which calls `CheckWeather`, which would have returned an error if the buffer length was incorrect. Try actually **reading** the entire 200k line codebase before you even **begin** to entertain the notion that I’m capable of a mistake.

PR #691, opened 2020-11-27
Add dummy file for convert

mtlynch:


parser/convert.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+
+
+
+  return 0;
+}
+
+int ConvertExcelDateToBasicDate(excelDate timestamp, basicDate* converted) {
```
TheWaySnrubThinks, line 38: This actually won’t work for January and February 1900.

mtlynch: Wow, nice catch!

PR #1076, opened 2023-07-20
Fix broken link to configuration.nix

brantonb:
There was a missing `/` in the URL.


mtlynch (approved):
Ah, looks like I missed this in #1073.

Thanks, Branton! LGTM

PR #1519, opened 2025-07-17
Fix motherboard typo

albertchae:


mtlynch:
Thanks for the fix, @albertchae!

